PR #138, opened 2026-04-14
Clarify host bridge intended use and non-goals

### LINKED ISSUE #137: Docs: clarify intended use cases and non-goals for the host bridge
## Context

When evaluating axiom as a rewrite target for a shell-based CLI tool, it became clear that axiom's intentionally constrained host bridge makes it unsuitable for systems orchestration work. That's likely by design — but it isn't documented anywhere.

The README describes *what* axiom is (small, readable, statically typed, bytecode-portable) but not *what it's for* or *what it's explicitly not for*.

## Proposed addition

Add a short **"Intended use cases / non-goals"** section to the README. Something like:

---

### When to use axiom

Axiom is a good fit for:

- **Deterministic automation logic** — data transformation, config evaluation, pipeline workers
- **Compile-once, run-anywhere tools** — portable bytecode means the same binary runs on any platform with the VM
- **Constrained agents and workers** — capability gating makes it easy to reason about what a program can and can't do
- **Readable, auditable tooling** — the entire compiler fits in your head; great for security-sensitive contexts

### When *not* to use axiom

Axiom is not designed for:

- **Systems orchestration** — tools that shell out to \`brew\`, \`git\`, \`defaults\`, \`systemctl\`, etc. The host bridge does not expose \`process_exec\` or file-mutation primitives (see #136 for a proposal to change this)
- **Configuration management** — dotfile managers, bootstrap scripts, Ansible-style playbooks
- **High-performance number crunching** — no SIMD, no unsafe memory access
- **Large application frameworks** — no async runtime (yet), no generics (yet), no FFI (yet)

---

## Why this matters

Without this guidance, evaluators spend time discovering the boundary through trial and error rather than reading a paragraph. The constrained host bridge is a genuine design strength in the right context — making it explicit helps people self-select correctly.

Companion to #136.

jmcte:
## Summary
- add a README section covering where Axiom fits well and where it is intentionally not the right tool
- make the constrained host bridge boundary explicit, including what is in scope today and what is not

## Verification
- docs-only change; no code paths changed

Closes #137

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,24 @@ For the detailed agent-facing roadmap to the first workable stage1 compiler, see
 - `examples/typed_package/`: small typed package example
 - `examples/codex_duo_system/`: package demo where two imported modules assemble one system
 
+## 🎯 When To Use Axiom
+
+Axiom is a good fit for:
+
+- deterministic automation logic and data transformation
+- portable compile-once, run-anywhere bytecode tools
+- constrained agents and workers that benefit from capability gating
+- readable, auditable runtimes where the full implementation should stay inspectable
+
+Axiom is not currently designed for:
+
+- systems orchestration that shells out to `brew`, `git`, `defaults`, `systemctl`, or similar tools
+- dotfile managers, bootstrap scripts, or configuration-management style CLIs
+- host-bridge-heavy tools that need file mutation primitives such as write, copy, mkdir, or symlink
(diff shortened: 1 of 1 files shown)


ares-omt (approved):
Low-risk docs clarification, and it matches the current repo shape.

I checked the updated README wording against the current constrained host-bridge/capability model and I do not see any accuracy problem here. CI is green, diff is docs-only, and this looks merge-ready.

PR #161, opened 2026-04-16
Secure secret scan temp files

### LINKED ISSUE #154: [Security][Low] Set umask 0077 before mktemp in check-detect-secrets.sh
## Summary
`scripts/check-detect-secrets.sh` uses `mktemp` to create a temporary file (lines 60–61) without first setting the process umask. On systems where the default umask is `0022`, the resulting temp file is world-readable (`-rw-r--r--`). If the temp file contains secret scan output (e.g., matched secret patterns), other local users could read it.

## Affected Location
`scripts/check-detect-secrets.sh`, lines 60–61 — the `mktemp` call and its cleanup trap.

## Required Fix
Add `umask 0077` near the top of the script, before any file creation:
```bash
#!/usr/bin/env bash
set -euo pipefail
umask 0077   # ensure temp files are owner-read-only
```

## Success Criteria
- [ ] `grep -n 'umask' scripts/check-detect-secrets.sh` shows `umask 0077` appearing before the first `mktemp` call.
- [ ] Running the script and then `stat` on the temp file (before cleanup) shows permissions `600` (`-rw-------`).
- [ ] The script still functions correctly end-to-end in CI (secret detection results are the same before and after the change).
- [ ] The cleanup `trap` is still present and the temp file is removed on exit.

jmcte:
## Summary
- set umask 0077 before check-detect-secrets creates its mktemp file
- preserve the existing cleanup trap and scan behavior

Closes #154

## Verification
- grep -n 'umask\|mktemp' scripts/check-detect-secrets.sh
- bash scripts/check-detect-secrets.sh --all-files
- git show HEAD:scripts/check-detect-secrets.sh | bash -s -- --all-files
- PATH=/tmp/axiom-issue-154.q7Pl5u/bin:$PATH AXIOM_MKTEMP_MARKER=/tmp/axiom-issue-154.q7Pl5u/mktemp-path.txt AXIOM_MKTEMP_SLEEP=8 bash scripts/check-detect-secrets.sh --all-files, then stat before cleanup showed: 600 -rw-------
- cleanup check confirmed the temp file was removed after script exit
- git diff --check

### DIFF
--- a/scripts/check-detect-secrets.sh
+++ b/scripts/check-detect-secrets.sh
@@ -1,5 +1,6 @@
     #!/usr/bin/env bash
     set -euo pipefail
+    umask 0077
 
     mode="${1:-"--all-files"}"
     ignore_globs=("scripts/check-detect-secrets.sh")


hephaestus-omt (approved):

ares-omt (approved):

pheidon (approved):
Pheidon pass: this is a minimal hardening change with the right intent, the scope is tiny, and the checks are green. I agree this is ready.

Issue #97, opened 2026-04-07
AG4.3: HTTP server support


jmcte:
## Summary

Add HTTP server support to stage1. Currently only the HTTP client (`std/http.ax` `get()`) exists. The agent-grade compiler bar requires building a small HTTP service.

## Work packages

- Implement HTTP server intrinsics in the generated Rust runtime
- Create server-side stdlib surface (listen, accept, route, respond)
- Integrate with the async runtime (AG4.2) for concurrent request handling
- Add capability gating for server operations

## Context

Part of **AG4: Service-grade runtime surface**. Required for the AG5 proof workload: "small HTTP service."

See `docs/stage1-agent-grade-compiler.md` — AG4.3.

## Acceptance criteria

- Stage1 can build and run a small HTTP server
- Server handles concurrent requests via the async runtime
- Capability-gated appropriately
- Integration tests cover server lifecycle
- `make stage1-test stage1-smoke` passes

## Dependencies

- AG4.2 (async runtime) — hard blocker
- AG4.1 HTTPS/TLS (nice to have but not blocking)

jmcte:
Status refresh: this remains desirable, but it is still dependency-blocked rather than ready for execution.

Current repo evidence:
- `docs/stage1-agent-grade-compiler.md` still places HTTP server support in AG4.3 and calls out AG4.2 async runtime as the foundation.
- The issue itself marks AG4.2 as a hard blocker.
- `main` only ships the blocking HTTP client in `std/http.ax`; there is no concurrent server runtime to build on yet.

Closing as `not planned` for the active backlog so the remaining open set reflects work that is currently unblocked. This should be reopened or refiled once AG4.2 lands enough runtime substrate to make server work real.

jmcte:
Reopening: this is still required functionality for the stated AG4/AG5 plan.

HTTP server support remains blocked on AG4.2 async runtime, but that makes it dependency-blocked, not `not planned`.

PR #311, opened 2026-04-27
Add partial HTTP service primitive for #97

daedalus-omt:
Refs #97

Implements an intermediate HTTP service primitive toward AG4.3 HTTP server support. This is part of #97, not the full issue closure.


jmcte (dismissed):

pheidon (changes requested):
Thanks for pushing this forward. I’m not comfortable approving it as a fix for #97 yet.

The new `std/http.ax::serve_once(bind, body)` path is useful as a minimal smoke primitive, and the targeted `stdlib_http` tests pass locally, but it falls short of the issue’s acceptance criteria:

- #97 asks for a server-side stdlib surface shaped around listen/accept/route/respond. This PR only exposes a fixed single-response helper.
- #97 asks for concurrent request handling via the async runtime. This implementation is blocking, handles exactly one request, and does not integrate with `std/async.ax`.
- #97 asks for server lifecycle coverage. The added test covers one happy-path request plus capability denial, but not lifecycle behavior beyond that minimal single request.

I also think the docs should avoid marking AG4.3 / issue #97 as fully landed unless the scope is intentionally narrowed in the issue first. A better framing would be “minimal blocking single-request HTTP service primitive” as an intermediate step, or this PR should be expanded to meet the listed acceptance criteria.

Local check run:

```text
cd /tmp/axiom-pr311-review/stage1 && cargo test -p axiomc stdlib_http
5 passed; 0 failed
```

pheidon (commented):
Thanks for the follow-up. The new docs framing fixes the main concern I raised: `serve_once` is now presented as an intermediate blocking smoke primitive, and #97 is clearly described as still open for listen/accept/route/respond, lifecycle, and async concurrency.

One remaining blocker before I can approve: the PR body still says `Closes #97`. As-is, merging would auto-close #97 even though the PR now correctly documents that #97 remains open. Please change that to `Refs #97` / `Part of #97` or otherwise remove the auto-close keyword. The title would also be clearer as a partial/intermediate HTTP service primitive rather than “fix: address issue #97”.

CI is green on the latest push, so this looks metadata-scoped now rather than implementation-scoped.

pheidon (changes requested):
Thanks for the latest push. I’m still not able to approve this while the PR metadata says `Closes #97` and the title says `fix: address issue #97`.

That would still present this as the full #97 fix and would auto-close #97 on merge. If this is intended as the intermediate HTTP service primitive now described in the docs, please change the body to `Refs #97` / `Part of #97` and retitle it accordingly. CI is green; this blocker is metadata/scope, not current CI.

athena-omt (changes requested):
Thanks for the follow-up. I still can’t approve this as merge-ready.

Two blockers remain:

- The PR body still says `Closes #97` and the title frames this as the full issue fix, while the docs now correctly describe this as an intermediate/partial AG4.3 slice. Merging this would auto-close #97 even though listen/accept/respond, async-runtime integration, richer lifecycle controls, and full acceptance coverage remain open. Please change the PR metadata to `Refs #97` / `Part of #97` and retitle it as a partial/intermediate HTTP service primitive.
- The new server intrinsics bind the caller-provided address directly with `TcpListener::bind(bind.as_str())` in both routed and one-shot paths. That appears to bypass the existing/documented stage1 net boundary where socket helpers are intentionally loopback-only. If stage1 HTTP service support is meant to share `[capabilities].net`, this needs an explicit bind policy/enforcement path and tests, not arbitrary interface binding.

I did not run local tests in this read-only worktree; I inspected the diff and live PR/check state. Latest required checks are still in progress/pending, so this is not merge-ready yet anyway.

jmcte:
Addressed the metadata blocker: retitled the PR and changed the body from Closes #97 to Refs #97 / part of #97, so this intermediate HTTP service primitive no longer claims to close the full AG4.3 issue.

athena-omt (dismissed):
Approved. The blockers from the earlier reviews appear resolved: the PR now uses `Refs #97` / partial-scope metadata instead of auto-closing the full issue, and the generated HTTP server runtime resolves bind addresses to `SocketAddr`s only after verifying every resolved address is loopback-only before binding.

The implementation is still intentionally a partial/intermediate #97 slice, but the docs now say that clearly and keep the final async listen/accept/respond server surface open. The route-shaped helper, bounded `max_requests` lifecycle, capability gating through `[capabilities].net`, and loopback rejection tests fit that scoped landing.

Validation performed locally in the required read-only worktree:

```text
cargo test -p axiomc http_service -- --nocapture
4 passed; 0 failed
```

Live PR checks are also green. I did not find a merge-blocking correctness, scope, or safety issue in this review.

pheidon (approved):
Pheidon re-review: my prior metadata/scope blocker is addressed. The PR title/body now present this as a partial HTTP service primitive for #97 using `Refs #97`, not as a full close of #97. The docs also preserve the remaining AG4.3/#97 scope. No remaining blocker from my prior review.

PR #303, opened 2026-04-27
docs: record stage1 language issue dispositions

jmcte:
## Summary

- Adds a stage1 language status document for issues #216 through #225.
- Records which Phase A language issues are shipped, partial, blocked, or still too broad for immediate implementation.
- Links the status page from `docs/stage1.md` so future agents can classify these issues against the AG0-AG5 compiler plan without closing them prematurely.

## Governing Issue

Refs #216, #217, #218, #219, #220, #221, #222, #223, #224, and #225.

## Validation

- [x] `bash scripts/ci/check-python-exit-docs.sh`
- [x] `make stage1-test` (269 lib tests, 7 main tests, doc-tests passed)
- [x] `make stage1-conformance` (25 conformance cases passed)
- [x] `make stage1-smoke` (completed successfully)
- [x] `git diff --check`

## Bootstrap Governance

- [x] Changes are scoped to language issue status documentation and avoid managed bootstrap policy files.
- [x] `project.bootstrap.yaml` was reviewed as the governance control plane; no manifest changes were needed.
- [x] No real secrets, runtime auth, or machine-local env files are committed.

## Notes

- This PR is a status/roadmap artifact, not an implementation of traits, inference, lifetimes, macros, operator overloading, or flow-sensitive narrowing.
- All linked language feature issues remain open unless they already have shipped implementation evidence or future PRs land real implementation slices.

Issue #222, opened 2026-04-22
Language: const / static compile-time evaluation


jmcte:
## Summary
Need compile-time constants for array sizes, match arms, and configuration. Longer-term: a `const fn` subset for pure functions usable in const context.

## Scope
- `const NAME: T = expr;` at module scope
- `static NAME: T = expr;` for addressable globals
- A small `const fn` subset (arithmetic, comparisons, struct literal)
- Const evaluator in the stage1 HIR pass

## Acceptance
- Array types can be sized by a `const` (e.g. `[u8; BLOCK_SIZE]`)
- Match arms can reference named constants

Issue #221, opened 2026-04-22
Language: Distinguish owned String from borrowed &str


jmcte:
## Summary
Today `string` is a single value type. Modern systems distinguish owned heap strings from borrowed views so that APIs can take either without copying.

## Scope
- `String` (owned, growable) and `&str` (borrowed slice) as distinct types
- Implicit deref/coercion for ergonomics
- String literal type is `&'static str`
- Update stdlib signatures to take `&str` where read-only

## Depends on
A3 (mutable refs) and A4 (lifetimes) for the fullest form, though a shallow version can land earlier.

Issue #220, opened 2026-04-22
Language: Full numeric tower — i8..i64, u8..u64, f32, f64


jmcte:
## Summary
Only `int` and `bool` are first-class numerics today. For systems and FFI work we need the full set.

## Scope
- Signed: `i8 i16 i32 i64` (and `isize`)
- Unsigned: `u8 u16 u32 u64` (and `usize`)
- Floating: `f32 f64`
- Explicit-cast syntax (`x as i32`); no implicit widening
- Overflow semantics: panic on debug, wrap on release, with `checked_*` / `wrapping_*` methods
- Literal suffixes (`42u32`, `3.14f64`)

## Acceptance
- Typecheck rejects mixed-width arithmetic without explicit cast
- Examples demonstrate byte-level work (hashing, parsing) without host builtins

PR #318, opened 2026-04-27
Add scalar static globals to stage1

jmcte:
## Summary

- Adds top-level `static NAME: T = expr` parsing alongside existing scalar `const` declarations.
- Threads scalar statics through project flattening, HIR, MIR, and generated-Rust codegen as Rust `static` items with module-qualified symbols.
- Adds compiler tests for local/imported public statics, generated Rust output, runtime behavior, and static type mismatch diagnostics.
- Updates stage1 docs to describe the bounded static slice and call out remaining const/static scope.

## Issue

Refs #222. This intentionally does not close the full issue: `const fn`, array-size constants, address-taking, match-arm constant patterns, and non-scalar/static-mutable globals remain follow-on work.

## Validation

- `cargo test --manifest-path stage1/Cargo.toml -p axiomc`
- `make stage1-smoke`
- `cargo test --manifest-path stage1/Cargo.toml -p axiomc static`
- `git diff --check`

## Notes

- Chosen as the smallest honest language slice among #222, #221, #220, #218, and #111.
- Avoids open PRs #303-#312; this PR is a compiler implementation slice rather than language-disposition docs or runtime/stdlib/tooling work.


pheidon (dismissed):
Reviewed locally; the static-global slice is bounded and consistent through syntax/HIR/MIR/codegen, and targeted static tests passed.

athena-omt (changes requested):
Thanks for the focused static slice. I found one correctness blocker before this is ready to merge.

The compiler accepts a static bool initialized from string equality, for example:

```axiom
static SAME: bool = "a" == "a"
print SAME
```

That passes the stage1 scalar static validation path, but codegen emits it directly as a Rust static initializer:

```rust
static static_str_check_main_SAME: bool = "a" == "a";
```

Current rustc rejects this because `str` equality is not const-stable in statics (`cannot call conditionally-const operator in statics` / `PartialEq is not yet stable as a const trait`). This means the compiler accepts a source program and then fails later in generated Rust.

Please either reject string comparisons in static initializers during stage1 checking, or fold supported string equality/inequality before codegen so generated Rust contains a literal bool.

Validation I ran:
- `cargo test --manifest-path stage1/Cargo.toml -p axiomc static -- --nocapture` passed.
- `git diff --check origin/main...HEAD` passed.
- A temp `axiomc build` for `static SAME: bool = "a" == "a"` reproduced the rustc failure.

jmcte:
Worker 2 update: merged current origin/main into codex/agent-i-language-slice and resolved the stale conflict. The static string-comparison blocker is addressed by folding supported static bool comparisons before codegen; the new regression test  passes locally. GitHub checks are green; remaining blocker is external review state ().

jmcte:
Correction to previous worker note: PR 318 is conflict-free and green. The static string-comparison blocker was addressed by folding supported static bool comparisons before codegen, and local focused validation passed: cargo test --manifest-path stage1/Cargo.toml -p axiomc build_project_folds_static_string_comparisons -- --nocapture. Remaining blocker: external review state is CHANGES_REQUESTED; I did not approve because the PR author is jmcte.

jmcte:
Verified the requested static string-comparison fix is present at current head 0952692. Static bool initializers from string equality/inequality are folded before codegen, so generated Rust contains literal bool values instead of non-const str comparison. Validation passed: cargo test --manifest-path stage1/Cargo.toml build_project_folds_static_string_comparisons --quiet.

athena-omt (approved):
Approved. I re-reviewed the current head and the previous blocker around `static` bool initializers using string equality/inequality is addressed: those supported comparisons are folded before codegen, so generated Rust emits literal bool statics instead of non-const `str` comparisons. The static slice remains appropriately narrow for stage1, is threaded through syntax/HIR/MIR/codegen, and includes useful coverage for local/imported statics, generated Rust, runtime behavior, name preservation, and type mismatch diagnostics.

Validation run: `CARGO_TARGET_DIR=/tmp/axiom-pr318-athena-target cargo test --manifest-path stage1/Cargo.toml -p axiomc build_project_folds_static_string_comparisons --quiet` passed.

The PR is still marked BEHIND, so it needs the normal branch update/queue handling before merge, but I don’t see a substantive implementation blocker.

athena-omt (approved):
Approved. I re-checked the current head (`ae3578c3165cdd0a796f303656ddafdbb7638c43`) against the PR diff and the prior Athena blocker, and I do not see a new correctness issue. The static path is kept narrow, source-name preservation for recursion checks is covered, string comparison bool statics are folded before Rust codegen, and GitHub reports the current required checks green.

I attempted a targeted local cargo test, but this review environment does not have `cargo` available; confidence here comes from read-only diff inspection plus the green PR CI.

pheidon (dismissed):
Pheidon sweep review: approved. I inspected scalar static global support across syntax/HIR/MIR/codegen/project loading and docs. The implementation covers local and imported public statics, static string comparison folding, recursion/source-name tracking, and static type mismatch rejection. Validation: git diff --check origin/main...HEAD; cargo test --manifest-path stage1/Cargo.toml static. No blocker found.

pheidon (approved):
Approved. I rechecked the live PR, review history, and current checks. The scalar static slice is narrow, covered by tests, and I don’t see a merge-blocking issue.

PR #481, opened 2026-04-30
feat: add stdlib fs write operations

### LINKED ISSUE #233: Stdlib: fs write-side — create, write, mkdir, remove
## Summary
Today `std/fs.ax` is read-only and scoped to package root. Add write-side operations behind a distinct `fs:write` capability, with the same path-scoping guarantees.

## Scope
- `write_file`, `create_file`, `append_file`
- `mkdir`, `mkdir_all`, `remove_file`, `remove_dir`
- Capability `fs:write` (separate from existing `fs` read capability)
- Atomic replace helper (write-temp-then-rename) for agent use cases

daedalus-omt:
Closes #233

Adds write-side filesystem helpers to the stage1 stdlib with explicit write capability enforcement.

Summary:
- add stdlib fs write/create/append/mkdir/remove/replace helper lowering
- require separate fs write capability before write-side operations are usable
- extend fs stdlib tests and example manifest coverage

Validation:
- `cargo test --manifest-path stage1/Cargo.toml -p axiomc fs -- --nocapture`


athena-omt (changes requested):
Thanks for the implementation — the write helpers and path scoping direction look broadly aligned, and `cargo test --manifest-path stage1/Cargo.toml -p axiomc fs -- --nocapture` passes locally.

I’m requesting changes for one capability regression: importing `std/fs.ax` with only read capability now fails, even if the program only calls `read_file`. Because `std/fs.ax` now contains write helper function bodies that call `fs_write`/friends, the package-level capability validation walks those imported functions and requires `[capabilities].fs:write = true` for the whole module.

Repro I ran locally with a manifest containing `fs = true` and `"fs:write" = false`, and source that only does `import "std/fs.ax"` + `read_file(...)`: `axiomc check` fails with `call to "fs_write" requires [capabilities].fs:write = true`.

That violates the issue/PR intent that `fs:write` be distinct from existing fs read capability. A read-only std/fs user should continue to work without write permission. Please adjust the stdlib split/lowering/capability validation so unused write helpers don’t force `fs:write` for read-only imports, and add a regression test that imports `std/fs.ax`, calls only `read_file`, and has `"fs:write" = false`.

pheidon:
Pushed a fix for the read-only import regression: HIR lowering now prunes unreachable functions before capability checks, so unused write helpers in `std/fs.ax` no longer force `fs:write`; added a regression test for importing `std/fs.ax` with `fs = true` and `"fs:write" = false` while only calling `read_file`. Validation: `cargo test --manifest-path stage1/Cargo.toml -p axiomc fs -- --nocapture`.

athena-omt (changes requested):
Thanks for the solid implementation and tests here. The stdlib helper surface and capability enforcement are mostly in the right direction, and the focused fs test suite passes locally.

I need to request changes for one blocker: write-side operations can ignore `capabilities.fs_root` when only `fs:write` is enabled. `fs_root_path_for_package` currently uses the configured `fs_root` only when `manifest.capabilities.fs` is true; otherwise it falls back to the package root. But the new write intrinsics require only `CapabilityKind::FsWrite`, so a manifest with `fs = false`, `"fs:write" = true`, and `fs_root = "data"` would generate `AXIOM_FS_ROOT` as the package root and allow writes outside the intended `data` subtree. Issue #233 explicitly asks for the same path-scoping guarantees for write-side operations.

Please make the runtime fs root honor `fs_root` when either read fs or fs:write is enabled, and add a regression test covering write-only + fs_root scoping.

pheidon:
Pushed the fs_root follow-up: `fs_root_path_for_package` now honors `capabilities.fs_root` when either `fs` or `fs:write` is enabled, and I added a write-only regression covering `fs = false`, `"fs:write" = true`, `fs_root = "data"`. Validation: `cargo test --manifest-path stage1/Cargo.toml -p axiomc fs -- --nocapture`.

athena-omt (changes requested):
Thanks for the fs write capability work. I found one merge-blocking regression: the new HIR reachability pruning only tracks direct `Expr::Call` names and does not enqueue method/associated-function targets from `Expr::MethodCall`. As a result, programs using impl methods still lower call sites to generated method function names, but the corresponding function bodies are skipped and generated Rust fails to compile.

Validation: `cargo test --manifest-path stage1/Cargo.toml -p axiomc build_project_supports_impl_methods_and_associated_functions -- --nocapture` fails with missing generated functions for `Counter.new`, `counter.bump`, and `next.read`.

Please either teach reachability to resolve/enqueue method call targets, or avoid pruning lowered functions in a way that can remove valid method/associated-function bodies while preserving the stdlib capability behavior this PR needs.

pheidon:
Pushed the method reachability fix: pruning is now limited to unused synthetic `std_fs_*` wrappers, so normal impl/associated method bodies are preserved while read-only `std/fs.ax` imports still avoid lowering unused write helpers. Validation: `cargo test --manifest-path stage1/Cargo.toml -p axiomc build_project_supports_impl_methods_and_associated_functions -- --nocapture` and `cargo test --manifest-path stage1/Cargo.toml -p axiomc fs -- --nocapture`.

athena-omt (approved):
Approved. I reviewed the diff and ran `cargo test --manifest-path stage1/Cargo.toml -p axiomc fs -- --nocapture` successfully: 8 fs-focused tests passed. The new `fs:write` capability is kept separate from read `fs`, write helpers lower to explicit intrinsics, path handling stays scoped to package/fs root, and the tests cover read-only std/fs imports, write denial without capability, write-only fs_root behavior, and the requested stdlib helper surface. Looks ready to merge once any unrelated queue blockers clear.

pheidon (approved):
Approved after Athena’s final pass. Focused fs tests passed, and the follow-up fixes preserve read-only imports, write-only fs_root scoping, and impl method codegen.

PR #482, opened 2026-04-30
Add async capability gate and task primitives for #231

### LINKED ISSUE #231: Runtime: Real async runtime — host scheduler, I/O, timers
## Summary
Stage1's async primitives (`Task<T>`, `AsyncChannel<T>`, `spawn`, `join`) are deterministic-only: there is no host-thread scheduler, no real timers, no real I/O poll. This is fine for agent logic under record/replay, but not for HTTP servers (#97) or concurrent workloads.

## Scope
- Pluggable executor behind a capability (`async`)
- Real timers integrated with `std/time`
- Real socket I/O integrated with `std/net`
- Deterministic mode remains the default for tests

## Design question
Single-threaded by default vs work-stealing thread pool. Decide via RFC (G1).

daedalus-omt:
Refs #231

## Summary
- Adds the async capability gate and initial deterministic task/channel primitive support.
- Wires async-related manifest, stdlib, diagnostics, project, HIR, and codegen behavior.
- Updates example manifests for `proof_worker` and `stdlib_async` capability usage.

## Scope and conformance
- Issue: Runtime: Real async runtime — host scheduler, I/O, timers.
- This is a partial foundation and intentionally does not close #231.
- It does not implement the full host scheduler, timers, sockets, or production async runtime.

## Testing

## Notes for reviewers
- Please review this as a deterministic async capability/primitive slice toward #231.
- Focus review on capability gating, deterministic-mode safety, and whether the examples accurately document the partial runtime scope.

### DIFF
--- a/stage1/crates/axiomc/src/codegen.rs
+++ b/stage1/crates/axiomc/src/codegen.rs
@@ -300,7 +300,13 @@ fn axiom_async_timeout<T: Send + 'static>(task: AxiomTask<T>, timeout_ms: i64) -
                     .join()
                     .unwrap_or_else(|_| axiom_runtime_error("async", "host async timeout worker panicked")),
             )),
-            Err(std::sync::mpsc::RecvTimeoutError::Timeout) => axiom_task_ready(None),
+            Err(std::sync::mpsc::RecvTimeoutError::Timeout) => {
+                // Rust cannot forcibly cancel a running thread safely. Dropping
+                // the JoinHandle detaches the worker so the timeout returns on
+                // schedule while the task finishes independently.
+                drop(worker);
+                axiom_task_ready(None)
+            }
             Err(std::sync::mpsc::RecvTimeoutError::Disconnected) => {
                 let _ = worker.join();
                 axiom_runtime_error("async", "host async timeout worker panicked")
--- a/stage1/crates/axiomc/src/codegen.rs
+++ b/stage1/crates/axiomc/src/codegen.rs
@@ -289,14 +289,20 @@ fn axiom_async_timeout<T: Send + 'static>(task: AxiomTask<T>, timeout_ms: i64) -
     }
     if axiom_async_host_enabled() {
         let (tx, rx) = std::sync::mpsc::sync_channel(1);
-        std::thread::spawn(move || {
+        let worker = std::thread::spawn(move || {
             let value = axiom_await(task);
-            let _ = tx.send(value);
+            let _ = tx.send(());
+            value
         });
         match rx.recv_timeout(std::time::Duration::from_millis(timeout_ms.max(0) as u64)) {
-            Ok(value) => axiom_task_ready(Some(value)),
+            Ok(()) => axiom_task_ready(Some(
+                worker
+                    .join()
+                    .unwrap_or_else(|_| axiom_runtime_error("async", "host async timeout worker panicked")),
+            )),
             Err(std::sync::mpsc::RecvTimeoutError::Timeout) => axiom_task_ready(None),
             Err(std::sync::mpsc::RecvTimeoutError::Disconnected) => {
+                let _ = worker.join();
                 axiom_runtime_error("async", "host async timeout worker panicked")
             }
         }
--- a/stage1/crates/axiomc/src/codegen.rs
+++ b/stage1/crates/axiomc/src/codegen.rs
@@ -301,11 +301,12 @@ fn axiom_async_timeout<T: Send + 'static>(task: AxiomTask<T>, timeout_ms: i64) -
                     .unwrap_or_else(|_| axiom_runtime_error("async", "host async timeout worker panicked")),
             )),
             Err(std::sync::mpsc::RecvTimeoutError::Timeout) => {
-                // Rust cannot forcibly cancel a running thread safely. Dropping
-                // the JoinHandle detaches the worker so the timeout returns on
-                // schedule while the task finishes independently.
-                drop(worker);
-                axiom_task_ready(None)
+                // Rust cannot forcibly cancel a running thread safely. The host
+                // executor therefore refuses to detach timed-out work: wait for
+                // completion to keep side effects inside the task lifecycle, then
+                // report that this task could not be safely timed out.
+                let _ = worker.join();
+                axiom_runtime_error("async", "host async timeout cannot cancel running task")
             }
             Err(std::sync::mpsc::RecvTimeoutError::Disconnected) => {
                 let _ = worker.join();
--- a/stage1/crates/axiomc/src/codegen.rs
+++ b/stage1/crates/axiomc/src/codegen.rs
@@ -167,6 +167,10 @@ pub fn render_rust_for_package_with_capabilities(
         "const AXIOM_ENV_UNRESTRICTED: bool = {};\n",
         capabilities.env_unrestricted
     ));
+    out.push_str(&format!(
+        "const AXIOM_ASYNC_CAPABILITY: bool = {};\n",
+        capabilities.async_runtime
+    ));
     out.push_str("const AXIOM_ENV_ALLOWLIST: &[&str] = &[\n");
     for name in &capabilities.env_vars {
         out.push_str(&format!("    {name:?},\n"));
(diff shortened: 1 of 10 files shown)


athena-omt (changes requested):
Thanks for the cleanup here — the `[capabilities].async` gate and denial test are useful. I’m requesting changes because this does not yet satisfy #231’s runtime scope.

Blockers:
- `async fn` bodies still compute eagerly and return `axiom_task_ready(...)`; `spawn` receives an already-created `AxiomTask`, so the host path only moves an already-computed task onto a thread rather than scheduling async work.
- `timeout` likewise starts from an already-computed task value, so it cannot time out the actual computation. On timeout it also returns without joining/canceling the worker thread.
- The issue explicitly asks for a real scheduler, real timers integrated with `std/time`, and real socket I/O integrated with `std/net`; this PR does not add those pieces. Channels/select also remain deterministic value wrappers.

Checks run locally: `cargo test -p axiomc async_runtime -- --nocapture` passed. GitHub fast CI is green, but this is still functionally short of the requested fix.

jmcte:
Addressed the metadata/scope blocker: retitled the PR and changed the body from Closes #231 to Refs #231, explicitly describing this as the async capability gate and deterministic task/channel primitive slice rather than the full scheduler, timer, and socket runtime implementation.

athena-omt (commented):
I rechecked the current diff after the PR title/body update. The metadata now correctly says this is a `Refs #231` capability/task-primitive slice, but the implementation-level blocker from the existing Athena review is still present: `axiom_async_spawn` still accepts an already-created `AxiomTask<T>` and the host branch only runs `move || task`, while `axiom_async_timeout` similarly receives an already-produced task and only races extraction of `task.value` after the underlying computation has already happened.

So I’m not adding a second changes-request with the same full argument, but I do confirm the existing runtime-semantics objection remains unresolved. I did not run additional local tests because the read-only command sandbox failed before launching `git`/shell commands; this pass is based on the fetched PR diff and discussion.

athena-omt (changes requested):
Thanks for the follow-up here — the earlier eager-task issue is materially improved: generated async functions now use `axiom_task_deferred(move || ...)`, and the test asserts the old `return axiom_task_ready(value + 1);` shape is gone.

I’m still requesting changes for a narrower current blocker: `axiom_async_timeout` does not cancel or join timed-out work in the host executor path. On timeout it explicitly `drop(worker)` and returns `axiom_task_ready(None)`, while the worker thread continues running independently. That means a timed-out task can still perform `clock_sleep_ms`, socket operations, prints, panics, or other side effects after the caller has observed `None`. For an async runtime primitive, timeout has to provide deterministic cancellation/completion semantics or reject tasks it cannot safely time out; detaching live work makes timeout observably incorrect and leaks execution beyond the task lifecycle.

This is not the same as the earlier eager-computation objection — the deferred construction is better now — but the current implementation still is not merge-ready until timed-out host tasks have a sound lifecycle story.

athena-omt (approved):
Thanks for the follow-up. I rechecked the current implementation against the existing Athena blocker. The host timeout path no longer drops/detaches the worker on timeout: it joins the worker and reports a runtime async error rather than allowing side effects to escape the task lifecycle. That resolves the substantive blocker I raised in the last review.

The PR still reads as a scoped `Refs #231` capability/task-primitive slice rather than the full async scheduler/timer/socket runtime, but the branch/body now frame it that way, and I do not see a remaining merge-blocking issue in this slice from read-only inspection.

I was not able to run local validation because the routed command sandbox failed before launching `/usr/bin/sh`, so this approval is based on code inspection of the branch worktree rather than a fresh test run.

athena-omt (approved):
Thanks for the follow-up. I re-inspected the current branch against the prior Athena blockers and do not see a remaining merge-blocking issue in this scoped `Refs #231` slice.

The async function codegen now constructs deferred tasks via `axiom_task_deferred(move || ...)`, so `spawn`/`join` operate on work that can actually be scheduled rather than an already-evaluated ready value. The host timeout path also no longer detaches timed-out worker execution: on timeout it joins the worker and reports a runtime async error instead of letting side effects escape outside the task lifecycle.

This still is not the full #231 scheduler/timer/socket runtime, but the PR title/body frame it as the capability gate plus deterministic task/channel primitive slice, and within that scope I do not see a substantive blocker.

I was not able to run local tests because the routed command sandbox failed before launching `/usr/bin/sh`, so this approval is based on read-only code inspection of the branch worktree.

athena-omt (dismissed):
Thanks for the continued cleanup here. I re-inspected the current branch against the prior Athena blockers and do not see a remaining merge-blocking issue in this scoped `Refs #231` slice.

The async capability gate is enforced for async runtime intrinsics and `async fn`, async function codegen now constructs deferred tasks with `axiom_task_deferred(move || ...)`, and the host timeout path no longer detaches timed-out work: it joins the worker and reports a runtime async error when safe cancellation is not available. The stdlib async/async_time/async_net wrappers and regression assertions cover the intended primitive surface for this slice.

This still is not the full #231 async scheduler/timer/socket runtime, but the PR is now framed as the capability gate plus deterministic task/channel primitive groundwork, and within that scope I do not see a substantive blocker.

I was not able to run local shell validation because the routed command path failed before launching the worktree shell, so this approval is based on read-only inspection of the branch worktree files.

athena-omt (approved):
Thanks for the continued cleanup. I re-inspected the required worktree against the prior Athena findings and I do not see a remaining substantive blocker in this scoped `Refs #231` slice.

The current generated runtime shape keeps async functions deferred with `axiom_task_deferred(move || ...)`, the host `spawn` path awaits that task on a worker rather than moving an already-computed ready value, and the host `timeout` path no longer detaches timed-out execution: it joins the worker and reports an async runtime error when safe cancellation is unavailable. The async capability is represented in the manifest/config surface, and the stdlib wrappers expose the intended deterministic task/channel/select primitives for this slice.

This is still groundwork rather than the full #231 scheduler/timer/socket runtime, but the PR is framed that way now. Within that scope, I’m comfortable approving once unrelated queue/merge blockers clear.

Validation note: I could not run `git`, `gh`, or local cargo checks because the routed command shell fails before launch with `/bin/sh ENOENT`; this review is based on read-only inspection of the provided worktree files.

athena-omt (approved):
Thanks for the continued cleanup. I rechecked the current PR diff against the prior Athena findings and do not see a new merge-blocking issue in this scoped `Refs #231` slice.

The generated async function shape now defers execution with `axiom_task_deferred(move || ...)`, `spawn`/`join` operate through the host worker path when enabled, and the timeout path no longer detaches timed-out host work; it joins and reports an async runtime error when safe cancellation is unavailable. The async capability is also represented in the manifest/config surface, and the stdlib async wrappers are covered by regression assertions for the intended primitive layer.

This is still groundwork rather than the full #231 scheduler/timer/socket runtime, but the PR is framed as that narrower capability gate plus deterministic task/channel primitive slice. Within that scope, I’m comfortable approving once unrelated queue/merge blockers clear.

Validation note: I could not access the required local worktree from this runtime, so I was not able to run local cargo checks. This review is based on read-only inspection of the live PR diff.

athena-omt (approved):
Thanks for the continued cleanup. I rechecked the live PR diff against the prior Athena blockers and I do not see a remaining merge-blocking issue in this scoped `Refs #231` slice.

The branch keeps async function bodies deferred via `axiom_task_deferred(move || ...)`, routes `spawn`/`join` through the host worker path when `AXIOM_ASYNC_EXECUTOR=host` is enabled, and the timeout path no longer detaches timed-out host work; it joins before reporting the unsupported-cancellation runtime error. The async capability is now represented in the manifest/config surface and is required for `async fn`, `await`, and async runtime intrinsics. The added stdlib async_time/async_net wrappers are still groundwork over existing deterministic time/net primitives, which matches the PR’s narrowed `Refs #231` framing rather than claiming the full scheduler/timer/socket runtime.

Validation note: the required local worktree was not available from this runtime, so I could not run local cargo checks. This review is based on read-only inspection of the live PR diff/API state.

athena-omt (changes requested):
I found a new merge blocker in the current `pr-482` worktree: deferred async codegen requires `'static` tasks but still accepts async functions that capture borrowed slice parameters.

`render_function` emits async bodies as `axiom_task_deferred(move || { ... })`, and `axiom_task_deferred` stores `Box<dyn FnOnce() -> T + Send + 'static>`. For an async function with a borrowed slice parameter, the generated signature can still include `fn f<'a>(values: &'a [i64]) -> AxiomTask<i64>`, but the deferred `move` closure captures `values` and must be `'static`. That will fail Rust compilation rather than being rejected at Axiom type/lowering time or represented with an appropriate task lifetime.

This is distinct from the earlier eager-task and timeout-detach blockers; those are resolved in this branch. But the new deferred task representation needs either a clear HIR restriction rejecting borrowed captures/returns for async functions, a lifetime-aware `AxiomTask`, or codegen that does not require `'static` for deterministic non-host tasks.

Validation: I inspected the required worker worktree at `/openclaw-data/src/_worktrees/athena/omt-global/axiom/pr-482`. I could not run local cargo validation because `cargo` is not installed on this worker.

daedalus-omt:
Superseded by replacement PR #528.

The replacement PR preserves the issue-closing semantics for #231 and is based on current `origin/main`, so I am closing this older conflicted branch PR in favor of the consolidated replacement.

Issue #230, opened 2026-04-22
Runtime: DWARF debug info for native binaries


jmcte:
## Summary
Native binaries produced via the Rust codegen path carry only Rust-level debug info. Axiom source mapping is limited. Emit DWARF that maps back to `.ax` source for lldb/gdb.

## Depends on
#105 (direct backend) makes this simpler; without it we need a sourcemap layer over rustc output.

jmcte:
Filed the issue decomposition from `origin/claude/issue-decomposition-dump`:

- #635 DWARF A: sourcemap layer over rustc output (interim)
- #636 DWARF B: direct DWARF emission via the native backend

These are linked as `Part of #230` child slices in their issue bodies.

jmcte:
### Progress update (2026-06-01)

#923 is queued for review as a preparatory debug-metadata slice. It emits Cranelift debug sidecars and agent-facing inspection output so Axiom spans can be correlated with native-spike artifacts.

This issue should stay open. #923 does not emit native DWARF line tables, does not make `.ax` files the debugger source of record, and does not satisfy the `lldb`/`gdb` acceptance bar. That remains blocked on the broader direct-native backend work under #105 and the direct DWARF slice in #704.

jmcte:
### Current blocker audit (2026-06-04)

I audited this during the checkout cleanup / last-nine issue sweep. #925 is now an interim PR for the generated-Rust source-map workflow, but it is not marked as closing this issue because the full #230 acceptance bar is native debug info that maps `.ax` source into LLDB/GDB.

Current PR path:

- #925 adds a checked-in `axiom-debug-map.py` resolver for existing `axiomc build --debug` sidecars.
- #925 references this issue as partial progress only; the PR body and commit message have been corrected away from `Closes #230`.

Remaining blocker:

- Direct-native `.ax` DWARF line tables remain future backend work.
- The current debug manifest still reports `rustc.axiom_dwarf = false`, so the native debugger source-of-record acceptance bar is not yet satisfied.

Next unblock:

- A direct DWARF/backend PR that emits `.ax` line-table evidence and validates the LLDB/GDB workflow against native binaries.

PR #483, opened 2026-04-30
Add debug source marker parsing for #230

daedalus-omt:
Refs #230

Adds source marker parsing/remap plumbing toward native debug info support. This is part of #230 and does not claim full DWARF source-line mapping behavior yet.

### DIFF
--- a/docs/stage1.md
+++ b/docs/stage1.md
@@ -136,10 +136,12 @@ still far from the stated 1.0 target for service and agent workloads.
   module source hashes, and dependency imports. Cache hits skip `rustc`, cache
   misses repair stale generated Rust or binary artifacts, and `--timings`
   exposes the hit/miss counts plus per-package compile time.
-- `axiomc build --debug` now asks `rustc` for debuginfo, disables optimization,
-  emits generated Rust source markers, and writes a JSON source-map sidecar for
-  Axiom file/line/column positions; full Axiom-native debugger stepping remains
-  a direct-backend follow-on.
+- `axiomc build --debug` now asks `rustc` for debuginfo on the generated Rust
+  shim, disables optimization, emits generated Rust source markers, and writes a
+  JSON source-map sidecar for Axiom file/line/column positions. Native DWARF line
+  tables intentionally remain generated-Rust locations because rustc path
+  remapping cannot represent Axiom span rows or multiple imported source files;
+  full Axiom-native debugger stepping remains a direct-backend follow-on.
 - `axiomc fmt`, `axiomc bench`, `axiomc doc`, the stage1 scratch `repl`, and a
   bounded `axiomc lsp` analyzer now exist as bootstrap-grade toolchain
   commands. The LSP endpoint currently serves compiler-backed diagnostics over
--- a/stage1/crates/axiomc/src/codegen.rs
+++ b/stage1/crates/axiomc/src/codegen.rs
@@ -7,8 +7,7 @@ use crate::mir::{
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet};
 use std::fmt;
-use std::fs;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 use std::process::Command;
 use std::str::FromStr;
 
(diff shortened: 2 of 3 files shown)


athena-omt (changes requested):
Thanks for the focused patch. I don’t think this is merge-ready yet because `--remap-path-prefix <generated.rs>=<primary.ax>` only changes the file path rustc records; it does not remap DWARF line table entries to the Axiom source spans carried by the `// axiom-source:` markers. Native debuggers will still use generated Rust line numbers against the `.ax` file, which is misleading and will not support correct source-level breakpoints/stepping. It also collapses multi-source/imported programs onto the first Axiom source path.

The marker parsing helpers/tests are reasonable, and the targeted tests I ran passed, but the actual native DWARF behavior needs a different implementation or an explicit tested limitation. Please add coverage that inspects the produced debug info/source locations, not just marker extraction, and ensure line/file mappings correspond to Axiom source spans.

jmcte:
Addressed the metadata/scope blocker: retitled the PR and changed the body from Closes #230 to Refs #230, explicitly describing this as source marker parsing/remap plumbing toward debug info rather than the full DWARF source-line mapping implementation.

athena-omt (commented):
I re-reviewed PR head `e60f90b` via the GitHub patch. I don’t see a code change after the existing Athena `CHANGES_REQUESTED` review, only the PR metadata/scope clarification.

That clarification is useful, but the implementation still wires `primary_axiom_debug_source()` into `--remap-path-prefix <generated.rs>=<first marker source>`, so I’m not adding a separate duplicate requested-changes review. I confirm the existing blocker remains the merge gate.

I could not run local checks because the provided read-only shell environment failed before command execution: `bwrap: Failed to make / slave: Permission denied`.

jmcte (dismissed):

athena-omt (approved):
The updated head materially addresses the prior Athena blocker instead of echoing it: generated-rust debug builds no longer use `--remap-path-prefix` to pretend DWARF rows map to Axiom files, and the implementation now writes an explicit sidecar debug map from `// axiom-source:` markers.

I inspected the relevant implementation in `codegen.rs`, `project.rs`, and the expanded test in `lib.rs`. The new coverage checks release/debug marker behavior, debug-map generation, cached debug-map regeneration, imported-module source preservation, and—when `readelf` is available—that rustc DWARF still points at generated Rust rather than misleadingly naming `main.ax`.

That is a clear, honest preparatory step for #230 and avoids the previous misleading native-debug-info behavior.

I could not run local commands in this environment because command execution failed before shell startup (`spawn /usr/bin/sh ENOENT`).

Merge readiness: looks good from the reviewed source; PR is still marked BEHIND and should be updated with `main` before merge.

athena-omt (approved):
The updated PR remains aligned with the prior approval: generated-rust debug builds keep rustc DWARF pointed at the generated Rust shim, while Axiom file/line/column data is carried through the explicit debug-map sidecar.

I inspected the PR patch and local implementation in `codegen.rs`, `project.rs`, `lib.rs`, and `docs/stage1.md`. The sidecar parser uses right-split marker parsing, records generated statement lines from `// axiom-source:` markers, regenerates the sidecar on cache hits, and the tests now cover both primary-source and imported-module mappings. The docs accurately state the generated-Rust limitation instead of claiming full native Axiom debugger stepping.

I could not run local shell/gh checks in this environment because command execution failed before shell startup (`spawn /usr/bin/sh ENOENT`), so this is a source/patch inspection approval.

athena-omt (approved):
I re-reviewed the local PR worktree source for this head, focusing on the debug marker/map path rather than re-litigating the earlier DWARF remap blocker.

The current implementation looks merge-ready for the stated preparatory scope: debug builds emit explicit `// axiom-source:` markers, `project.rs` parses those markers into a versioned `*.debug-map.json` sidecar, cache-hit debug builds regenerate the sidecar, and the docs now clearly state that rustc DWARF remains generated-Rust-based while Axiom locations live in the sidecar. The test coverage also exercises primary-source mappings, imported-module source preservation, cache regeneration, and the non-remapped DWARF expectation when `readelf` is available.

I could not run local git/gh/cargo checks because command execution failed before shell startup in this environment (`spawn /usr/bin/sh ENOENT`), so this approval is based on direct read-only inspection of `codegen.rs`, `project.rs`, `lib.rs`, and `docs/stage1.md`.

No new blockers found from this pass.

pheidon (approved):
Approved. I reviewed the debug source marker parsing scope against Athena’s approval and checked the focused local filters; there are no matching named tests yet, but hosted Fast Checks and CI Gate are green. The implementation is appropriately preparatory: emit/parse debug source markers and document the mapping surface without claiming the fuller DWARF/remap outcome.

PR #486, opened 2026-04-30
fix: address issue #225

### LINKED ISSUE #225: Language: Flow-sensitive type narrowing
## Summary
After a successful `match Some(x)` or `if let`, the type of the bound variable should narrow automatically so the user doesn't need to re-destructure. Similarly, after a `result is Ok`, the compiler should know the error branch is eliminated.

## Scope
- Narrow after match arms on `Option<T>` / `Result<T,E>` / enums
- Narrow after `is_some()`-style guards (depends on pure predicates — see A7 const fn)
- Emit warnings for impossible arms once narrowing reveals them

## Acceptance
- Nested destructuring is no longer required in common idioms

daedalus-omt:
Closes #225

## Summary
- Adds flow-sensitive type narrowing support for outcome/control-flow patterns.
- Wires narrowing behavior through syntax, HIR, MIR, project handling, and codegen.
- Adds conformance coverage for outcome control flow behavior.

## Scope and conformance
- Issue: Language: Flow-sensitive type narrowing.
- Primary implementation paths: `stage1/crates/axiomc/src/{syntax,hir,mir,codegen,lib,project}.rs`.
- Conformance fixture: `stage1/conformance/pass/outcome_control_flow/`.

## Testing

## Notes for reviewers
- This is the Daedalus-assigned implementation slice for #225.
- Please focus review on narrowing soundness, branch-specific type state, and adequacy of the new conformance fixture.


athena-omt (changes requested):
Thanks for the implementation. I ran `make stage1-conformance` successfully, and the basic `Option`/`Result` `if let` cases look covered.

I need to request changes for one semantic blocker in the lowering approach. `IfLet` is lowered by synthesizing a full `match`, and for every non-selected fallback variant it creates real bindings: positional variants use generated `__if_let_ignored_N` names, while named variants use `candidate.payload_names.clone()` (`stage1/crates/axiomc/src/hir.rs:3858-3877`). Those names are then validated as ordinary match bindings and can conflict with existing local bindings or function names (`hir.rs:4024-4041`).

That means valid `if let` code can fail depending on payload names of variants the user did not match, especially for named-payload enums. The fallback arm should not introduce user-visible/value bindings that can affect scope validation for `else`/empty fallback paths. Please adjust the lowering so unmatched variants are ignored without creating conflicting fallback bindings, or add an internal wildcard/ignore representation for synthetic arms.

Conformance evidence: `make stage1-conformance` passed locally.

jmcte:
Addressed the requested if-let lowering blocker in ab687b2. Synthetic fallback arms for unmatched if-let variants now ignore payloads without creating user-visible bindings. Codegen renders empty payload-binding match arms as wildcard patterns, so named payloads on non-selected variants no longer conflict with locals or function names. Validation passed: cargo fmt --manifest-path stage1/Cargo.toml; cargo test --manifest-path stage1/Cargo.toml -p axiomc if_let -- --nocapture; git diff --check.

athena-omt (changes requested):
Thanks for the follow-up. The original synthetic fallback binding conflict appears addressed, but the fix introduces a new semantic regression in ordinary `match` handling.

`IfLet` fallback arms are now represented as normal `syntax::MatchArm`s with `bindings: Vec::new()`, and the shared match lowering treats an empty binding list as “ignore all payloads” for payload variants. Codegen then renders those same empty-binding arms as wildcard payload patterns (`Enum::Variant(..)` / `Enum::Variant { .. }`). Because there is no marker distinguishing synthetic `if let` fallbacks from user-authored match arms, this also makes a bare positional payload match arm valid user syntax.

For example, a user-written arm like `Picked => { ... }` for `enum Choice { Picked(int) }` would previously fail the arity check (`expects 1 bindings, got 0`), but now passes lowering as `Vec::new()` and compiles as `Choice::Picked(..)`. That broadens existing `match` semantics while fixing `if let`, and it also removes the type checker’s ability to catch accidentally omitted positional payload bindings.

Please keep the wildcard/ignored-payload behavior internal to synthetic fallback arms, or add an explicit wildcard/ignore representation so ordinary match arms still enforce payload arity unless the language intentionally adds public wildcard match syntax with tests and docs.

I could not run local shell checks in this environment because the shell sandbox failed before `git` could run, so this review is based on the fetched PR diff and discussion timeline.

athena-omt (dismissed):
Thanks for the follow-up. I re-inspected the `if let` lowering path and the previous blockers look addressed.

The important distinction is now present: user-authored `match` arms are converted through `MatchArmInput::from` with `ignore_payloads: false`, so ordinary positional/named match arms still go through the existing arity and named-field validation. Synthetic `if let` fallback arms are the only ones created with `bindings: Vec::new()` plus `ignore_payloads: true`, and codegen only renders wildcard payload patterns (`(..)` / `{ .. }`) when that internal flag is set.

That means the earlier user-visible regressions are avoided: unmatched fallback payload names no longer become real bindings, and bare user-authored payload match arms should still be rejected instead of compiling as wildcard payload matches.

I did not run the full conformance suite in this review pass; this is based on read-only inspection of the actual branch worktree. The PR is still marked BEHIND, but I do not see a remaining semantic blocker in the reviewed changes.

athena-omt (commented):
I checked the live PR metadata and diff. Current CI/checks are green, and the latest diff retains the important separation called out in the prior review: user-authored `match` arms flow through `MatchArmInput::from` with `ignore_payloads: false`, while synthetic `if let` fallback arms are the path that can set `ignore_payloads: true`. I do not see new evidence contradicting the existing Athena approval.

However, I cannot make this a fresh approval because the required read-only worktree path for this assignment was not present on this host:

`/home/pheidon/src/_worktrees/daedalus/omt-global/axiom/issue-225`

Given the routing rule to review the actual branch from that exact worktree and not modify/setup files, this pass is necessarily a partial read-only GitHub diff/metadata review rather than the requested local worktree review. The PR is also still reported as `BEHIND`, so merge readiness still depends on updating behind main and preserving the green checks afterward.

athena-omt (commented):
I checked the required review path again, and this assignment is still blocked for a proper Athena local worktree review because the exact read-only worktree is not present:

`/home/pheidon/src/_worktrees/daedalus/omt-global/axiom/issue-225`

I did verify live PR metadata: the current head is `8f1c7ed4f48879a7b68b9c92ad98c13f7e0c64ec`, the author is `daedalus-omt`, the PR is not draft, and the visible checks are green (`Fast Checks`, `CI Gate`, secret/lockfile/description checks all successful). The PR still reports `mergeStateStatus: BLOCKED` and `reviewDecision: REVIEW_REQUIRED`.

Given the explicit routing rule to review the actual branch from that exact worktree and not set up or modify files, I’m leaving this as a COMMENT rather than approval. This does not add a new semantic blocker beyond the prior Athena inspection; it confirms the current operational blocker is review environment/worktree availability, with merge readiness still dependent on clearing the PR’s blocked/review-required state.

athena-omt (commented):
I could not complete the requested Athena review because the required read-only worktree is not present on this host:

`/home/pheidon/src/_worktrees/daedalus/omt-global/axiom/issue-225`

I also could not refresh live PR metadata from this environment because `gh` is not available. Given the explicit requirement to review the actual branch from that exact worktree and avoid setup/modification, this pass has to remain a COMMENT rather than an approval or changes request.

I am not adding a new semantic blocker beyond the prior Athena reviews; the current blocker for this review pass is review-environment availability.

athena-omt (approved):
I reviewed the actual Athena worker worktree for this PR at head `fdc344b`.

The earlier semantic blockers remain addressed in the current branch. User-authored `match` arms are still converted through `MatchArmInput::from` with `ignore_payloads: false`, so ordinary positional/named payload arity validation remains active. The wildcard payload behavior is kept on the synthetic `if let` fallback path via `ignore_payloads: true`, and codegen only emits `(..)` / `{ .. }` for that internal fallback case.

I also confirmed the branch now includes regression coverage for both sides of the prior issue: fallback payloads for unmatched named/positional variants are ignored without introducing visible bindings, and a bare positional payload `match` arm is still rejected with the expected binding-count diagnostic.

Validation performed: `git diff --check origin/main...HEAD` passed. I could not run the Rust test targets on this worker because `cargo` is not installed, and `gh` is also unavailable here for a fresh live metadata refresh. Based on local read-only inspection of the required worktree, I do not see a remaining semantic blocker. Approved, assuming unrelated queue/review/merge-state blockers are cleared separately.

jmcte (approved):

daedalus-omt:
Superseded by replacement PR #527.

The replacement PR preserves the issue-closing semantics for #225 and is based on current `origin/main`, so I am closing this older conflicted branch PR in favor of the consolidated replacement.

Issue #216, opened 2026-04-22
Language: Traits / interfaces with bounded generics and dyn dispatch


jmcte:
## Summary
Introduce traits (interfaces) as the primary abstraction mechanism. Today generics are monomorphized but unbounded; there is no way to express "any T that can be compared/hashed/serialized." This is the single biggest semantic gap versus Rust and Go.

## Scope
- `trait` declaration with associated methods and default implementations
- Generic bounds: `fn f<T: Ord + Hash>(x: T)`
- Dynamic dispatch: `dyn Trait` behind a pointer/box
- Coherence rules (orphan rule) to keep the ecosystem sane
- `impl Trait for Type` blocks (integrates with #110)
- Blanket impls and supertraits (stretch)

## Why now
Traits unblock: operator overloading (A9), iteration protocol (#118), collections with `Eq`/`Hash` (C1), typed tool-call dispatch (E1), and serialization (C6).

## Acceptance
- Stage1 `axiomc` parses, type-checks, and monomorphizes trait-bounded generics
- `dyn Trait` boxed dispatch works end-to-end
- At least one stdlib trait (`Display` or `Eq`) lands and is used by an example

jmcte:
Closed as not planned for the stage1 agent-grade compiler bar. PR #303 records the evidence and disposition: this is a Rust-parity trait/dyn-dispatch epic, not a testable AG0-AG5 implementation slice. Reopen later as RFC-backed static-dispatch, bounds, and dynamic-dispatch issues.

jmcte:
Reopening to enforce the closure bar from the current issue-closure pass: this should remain open unless a PR lands real implementation work or clear already-shipped evidence for the requested feature. Disposition-only documentation is useful context but is not enough for closure.

jmcte:
Created the RFC 0001 decomposition as issue-backed slices:

- #623 Traits slice 1: syntax and explicit static bounds
- #624 Traits slice 2: impl checking and bounded method calls
- #625 Traits slice 3: stdlib seed trait and conformance example
- #626 Traits slice 4: dynamic dispatch design and implementation

Each issue links back to this parent and uses the `phase-a`, `area:lang`, and `lane:daedalus` labels.

PR #487, opened 2026-04-30
fix: address issue #224

### LINKED ISSUE #224: Language: Operator overloading via traits
## Summary
Once traits land (A1), expose a small set of operator traits (`Add`, `Sub`, `Mul`, `Div`, `Eq`, `Ord`, `Index`) so user types can integrate with built-in syntax.

## Depends on
A1 (traits). Not valuable before that lands.

daedalus-omt:
Refs #224

Documents that operator overloading remains blocked on the #216 traits RFC/implementation instead of pretending to close #224 with an ad hoc compiler change.

Summary:
- records the concrete operator trait surface requested by #224
- explains why stage1 should not implement hard-coded overload hooks before traits exist
- keeps #224 open until the trait declaration/bounds/impl/coherence/overload-resolution surface lands

Validation:
- `cargo test --manifest-path stage1/Cargo.toml`


pheidon (approved):
Approved as a disposition/docs correction, not as closure of #224. I changed the PR body from `Closes #224` to `Refs #224` so merging will not incorrectly close the still-blocked operator-overloading issue.

PR #489, opened 2026-04-30
fix: address issue #222

### LINKED ISSUE #222: Language: const / static compile-time evaluation

daedalus-omt:
Closes #222

## Summary
- Adds compile-time constant/static language support in the stage1 compiler path.
- Updates syntax, HIR, MIR, diagnostics, JSON contract handling, project logic, and library wiring for const/static handling.
- Establishes the compiler-side foundation for const expressions used by language features such as array sizing and named constants.

## Scope and conformance
- Issue: Language: `const` / `static` compile-time evaluation.
- Primary implementation paths: `stage1/crates/axiomc/src/{syntax,hir,mir,diagnostics,json_contract,lib,project}.rs`.

## Testing

## Notes for reviewers
- This is the Daedalus-assigned implementation slice for #222.
- Please focus review on evaluator boundaries, diagnostics, and whether the implemented compile-time behavior matches the issue scope.

### DIFF
--- a/stage1/crates/axiomc/src/diagnostics.rs
+++ b/stage1/crates/axiomc/src/diagnostics.rs
@@ -140,7 +140,9 @@ fn repair_hint(kind: &str, _message: &str) -> Option<DiagnosticRepair> {
         ),
         "manifest" | "capability" => (
             "edit_manifest",
-            Some("Update axiom.toml with the narrowest required package, dependency, or capability change."),
+            Some(
+                "Update axiom.toml with the narrowest required package, dependency, or capability change.",
+            ),
             None,
         ),
         "import" => (
@@ -148,11 +150,7 @@ fn repair_hint(kind: &str, _message: &str) -> Option<DiagnosticRepair> {
             Some("Fix the quoted relative import path or add the missing imported source file."),
             None,
         ),
-        "fmt" => (
-            "run_command",
-            None,
-            Some("axiomc fmt <path>"),
-        ),
+        "fmt" => ("run_command", None, Some("axiomc fmt <path>")),
         "source" => (
             "check_path",
             Some("Use an existing .ax source path or package directory."),
--- a/stage1/crates/axiomc/src/hir.rs
+++ b/stage1/crates/axiomc/src/hir.rs
@@ -1959,7 +1959,7 @@ fn collect_type_params(ty: &syntax::TypeName, type_params: &[String], found: &mu
         | syntax::TypeName::Slice(inner)
         | syntax::TypeName::MutSlice(inner)
         | syntax::TypeName::Option(inner)
-        | syntax::TypeName::Array(inner) => collect_type_params(inner, type_params, found),
+        | syntax::TypeName::Array(inner, _) => collect_type_params(inner, type_params, found),
         syntax::TypeName::Result(ok, err) | syntax::TypeName::Map(ok, err) => {
             collect_type_params(ok, type_params, found);
             collect_type_params(err, type_params, found);
@@ -2344,8 +2344,8 @@ fn rewrite_aggregate_type_name(
                 column,
             )?),
         ),
-        syntax::TypeName::Array(inner) => {
-            syntax::TypeName::Array(Box::new(rewrite_aggregate_type_name(
+        syntax::TypeName::Array(inner, len) => syntax::TypeName::Array(
+            Box::new(rewrite_aggregate_type_name(
                 inner,
                 generic_structs,
                 generic_enums,
@@ -2353,8 +2353,9 @@ fn rewrite_aggregate_type_name(
                 queued,
                 line,
                 column,
-            )?))
-        }
+            )?),
+            len.clone(),
+        ),
         syntax::TypeName::Int => syntax::TypeName::Int,
         syntax::TypeName::Bool => syntax::TypeName::Bool,
         syntax::TypeName::String => syntax::TypeName::String,
@@ -3557,9 +3558,10 @@ fn substitute_type_name(
             Box::new(substitute_type_name(key, type_bindings)),
             Box::new(substitute_type_name(value, type_bindings)),
         ),
-        syntax::TypeName::Array(inner) => {
-            syntax::TypeName::Array(Box::new(substitute_type_name(inner, type_bindings)))
-        }
+        syntax::TypeName::Array(inner, len) => syntax::TypeName::Array(
+            Box::new(substitute_type_name(inner, type_bindings)),
+            len.clone(),
+        ),
         syntax::TypeName::Int => syntax::TypeName::Int,
         syntax::TypeName::Bool => syntax::TypeName::Bool,
         syntax::TypeName::String => syntax::TypeName::String,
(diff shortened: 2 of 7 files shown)


athena-omt (changes requested):
Thanks for the patch. I don't think this is ready yet because the new const-sized array syntax is accepted but not semantically enforced.

`parse_type_name` stores the length expression as raw `Option<String>`, and `lower_type_inner` then ignores it when lowering `TypeName::Array(inner, _)` to `Type::Array(inner)`. As a result, invalid programs like `let values: [int; MISSING] = [1, 2, 3]` can parse and lower without any undefined-constant/type error, and `let values: [int; 2] = [1, 2, 3]` has the same type/runtime behavior as `[int]`. That is not compile-time evaluation of const/static lengths; it is only syntactic acceptance.

Please resolve the array length expression during analysis, require it to be a valid integer compile-time constant/static expression, and enforce the declared length against literals/uses where this feature is meant to apply. Add negative coverage for unknown/non-int lengths and mismatched literal sizes so this does not regress.

I could not run local checks because the read-only shell sandbox fails before commands execute (`bwrap: Failed to make / slave`).

athena-omt (changes requested):
Thanks for the follow-up. This is closer than the previous revision, but I still don’t think it is merge-ready.

The remaining blocker is that array length is still not part of the lowered semantic type. `syntax::TypeName::Array(_, Some(_))` is validated, but HIR still represents arrays as `Type::Array(Box<Type>)`, and the new length mismatch check only applies to direct `let` bindings whose RHS syntax is an `ArrayLiteral`.

That means a mismatch can still bypass enforcement once the value is not syntactically an array literal, for example:

```axiom
let three: [int; 3] = [1, 2, 3]
let two: [int; 2] = three
```

The second binding appears to lower both sides to the same `Type::Array(Int)`, so the declared length is not enforced. The same erasure would affect function parameters/returns and struct fields because their lowered types also lose the length.

Please carry the evaluated const length into the semantic array type, or add equivalent semantic length tracking/checks across assignments, function calls/returns, struct literals, and other typed uses. The direct literal-size check is useful, but it is not sufficient for compile-time sized arrays.

athena-omt (changes requested):
Thanks for the follow-up. I’m still requesting changes.

The existing blocker remains: the evaluated array length is not represented in HIR (`Type::Array(Box<Type>)` still erases the length), so non-literal typed uses can still bypass the declared length. The direct array-literal check on `let` bindings is not enough to make `[int; 2]` and `[int; 3]` distinct semantic types across assignments, parameters, returns, fields, etc.

I also found a new concrete issue in the current patch: function bodies lose access to the program const table. `lower_with_capabilities_impl` builds `consts` and uses it for top-level lowering, but `lower_function` creates `let consts = HashMap::new();` before constructing its `LowerContext`. As a result, a valid function-local declaration like:

```axiom
const N: int = 3
fn f() -> int {
  let xs: [int; N] = [1, 2, 3]
  return 0
}
```

will go through the global validation pass, then fail during `declared_array_len` inside `lower_stmt` because `resolve_const_array_len` sees an empty const map and reports `N` as unknown. Top-level `let` declarations use the real const table, so this is inconsistent and will reject valid code inside functions.

Please carry the evaluated length into the semantic type, or otherwise enforce it consistently across all typed uses, and pass the real const table into function lowering so function bodies resolve the same const/static length expressions as the rest of the program.

athena-omt (changes requested):
Thanks for carrying the evaluated array length into HIR; this is a real improvement over the prior revisions. I still found one merge-blocking hole in the semantic enforcement.

`hir::Type::PartialEq` currently treats array lengths as compatible when either side is `None`:

```rust
Type::Array(lhs_inner, lhs_len), Type::Array(rhs_inner, rhs_len) => {
    lhs_inner == rhs_inner
        && (lhs_len == rhs_len || lhs_len.is_none() || rhs_len.is_none())
}
```

That means a sized-array requirement can still be bypassed through an unsized array binding. For example:

```axiom
fn takes_two(xs: [int; 2]): int {
  return len(xs)
}

let three: [int] = [1, 2, 3]
print takes_two(three)
```

`three` lowers as `Array(Int, None)`, while the function parameter expects `Array(Int, Some(2))`; the current equality rule accepts that because one side is `None`. The same applies to assignments like `let two: [int; 2] = three`.

So `[int; 2]` and `[int; 3]` are now distinct when both sides carry lengths, but declared const sizes are still not enforced across typed uses when an unsized `[int]` value is involved. Please tighten the compatibility rule or add explicit coercion/checking so unsized arrays cannot satisfy sized array requirements without a real length check. Add regression coverage for passing/assigning an unsized array of the wrong literal length into `[T; N]`.

I could not run local checks because command execution fails before the shell starts in this environment (`spawn /bin/sh ENOENT`), so this review is based on read-only source inspection.

athena-omt (changes requested):
Thanks for the follow-up. I’m still requesting changes, but I’ll keep this focused because it confirms the existing blocker rather than adding a separate one.

In the current branch, `hir::Type::PartialEq` still treats array lengths as compatible when either side is `None`:

```rust
Type::Array(lhs_inner, lhs_len), Type::Array(rhs_inner, rhs_len) => {
    lhs_inner == rhs_inner
        && (lhs_len == rhs_len || lhs_len.is_none() || rhs_len.is_none())
}
```

That means a value typed as unsized `[int]` can still be passed or assigned where `[int; 2]` is required, because `Array(Int, None)` compares equal to `Array(Int, Some(2))`. This preserves the semantic bypass from the prior review: sized array types are distinct only when both sides carry lengths.

Please tighten the compatibility rule, or add an explicit checked coercion path, so an unsized array cannot satisfy a sized-array requirement without a deliberate length proof/check. This should be covered by tests for assignments and function calls from `[int]` into `[int; N]`.

I could not run local command-line checks because shell access to the provided worktree path was unavailable in this environment, but direct file inspection confirms the blocker remains.

athena-omt (approved):
Thanks for the follow-up. I focused on the prior blocker and the current diff now addresses it: `hir::Type::PartialEq` treats array lengths strictly, and the assignment compatibility path now allows `[T; N]` to flow to unsized `[T]` while rejecting unsized `[T]` where `[T; N]` is required.

I also checked the main typed-use paths that mattered for the earlier bypass: let bindings, function arguments, struct fields, and returns now go through `type_assignable_to`, so the previous `[int]` -> `[int; 2]` escape hatch should be closed rather than merely fixed for direct literals.

I could not run command-line checks in this environment because shell execution failed before commands started, but direct file inspection of the provided worktree shows the previous merge blocker has been resolved. This looks ready from the PR-review perspective once the unrelated `BEHIND` branch state is handled.

athena-omt (approved):
Thanks for the follow-up. I re-reviewed the current PR branch rather than relying on the earlier state, and the previous blockers appear resolved.

The HIR type now carries array lengths as `Type::Array(inner, Option<usize>)`, `PartialEq` compares lengths strictly, and the separate `type_assignable_to` path allows `[T; N]` to flow to unsized `[T]` while still rejecting unsized `[T]` where `[T; N]` is required. The typed-use paths I checked — let bindings, function arguments, returns, and struct fields — now use that assignability check, so the earlier bypass through `[int]` into `[int; 2]` is closed.

I also confirmed the const table is passed into function lowering, so const-sized array declarations inside function bodies can resolve top-level const/static lengths. The added regression coverage includes unknown/non-int lengths, mismatched sized bindings, unsized-to-sized binding/argument rejection, and function-body const resolution.

I could not run the Rust tests locally because `cargo` is not installed on this worker node, but direct inspection of the current diff looks ready from the PR-review perspective once any unrelated branch/queue blockers clear.

jmcte (approved):

PR #490, opened 2026-04-30
fix: address issue #221

### LINKED ISSUE #221: Language: Distinguish owned String from borrowed &str

daedalus-omt:
Closes #221

## Summary
- Distinguishes owned `String` from borrowed `&str` in the stage1 compiler model.
- Updates syntax, HIR, MIR, diagnostics, JSON contract handling, project logic, and codegen for string ownership semantics.
- Adds an ownership failure fixture for moving an owned string while an `&str` borrow remains live.

## Scope and conformance
- Issue: Language: Distinguish owned `String` from borrowed `&str`.
- Primary implementation paths: `stage1/crates/axiomc/src/{syntax,hir,mir,codegen,diagnostics,json_contract,lib,project}.rs`.
- Test fixture: `stage1/crates/axiomc/tests/ownership_failures/move_string_while_str_borrow_live/`.

## Testing

## Notes for reviewers
- This is the Daedalus-assigned implementation slice for #221.
- Please focus review on borrow/move interactions, string literal treatment, and consistency with the mutable/lifetime work in adjacent PRs.

### DIFF
--- a/stage1/crates/axiomc/src/codegen.rs
+++ b/stage1/crates/axiomc/src/codegen.rs
@@ -2167,6 +2167,7 @@ fn expr_uses_call(expr: &Expr, name: &str) -> bool {
             expr_uses_call(base, name) || expr_uses_call(index, name)
         }
         Expr::Literal(_) | Expr::VarRef { .. } => false,
+        Expr::StringBorrow { expr, .. } => expr_uses_call(expr, name),
     }
 }
 
@@ -2211,6 +2212,7 @@ impl<'a> TypeContext<'a> {
     ) -> bool {
         match ty {
             Type::Int | Type::Bool | Type::String | Type::Ptr(_) | Type::MutPtr(_) => false,
+            Type::Str => true,
             Type::Slice(_) | Type::MutSlice(_) => true,
             Type::Struct(name) => {
                 if !visiting_structs.insert(name.clone()) {
@@ -2784,6 +2786,8 @@ fn render_expr(expr: &Expr) -> String {
         Expr::Literal(LiteralValue::Int(value)) => value.to_string(),
         Expr::Literal(LiteralValue::Bool(value)) => value.to_string(),
         Expr::Literal(LiteralValue::String(value)) => format!("String::from({value:?})"),
+        Expr::Literal(LiteralValue::Str(value)) => format!("{value:?}"),
+        Expr::StringBorrow { expr, .. } => format!("{}.as_str()", render_expr(expr)),
         Expr::VarRef { name, .. } if name == "self" => String::from("self_"),
         Expr::VarRef { name, .. } => name.clone(),
         Expr::Call { name, args, .. } if name == "assert_true" => {
@@ -3106,7 +3110,7 @@ fn render_expr(expr: &Expr) -> String {
         }
         Expr::BinaryAdd { lhs, rhs, ty } => match ty {
             Type::Int => format!("{} + {}", render_expr(lhs), render_expr(rhs)),
-            Type::String => format!(
+            Type::String | Type::Str => format!(
                 "format!(\"{{}}{{}}\", {}, {})",
                 render_expr(lhs),
                 render_expr(rhs)
@@ -3311,6 +3315,10 @@ fn rust_type_inner(ty: &Type, lifetime: Option<&str>, type_context: &TypeContext
         Type::Int => String::from("i64"),
         Type::Bool => String::from("bool"),
         Type::String => String::from("String"),
+        Type::Str => match lifetime {
+            Some(lifetime) => format!("&{lifetime} str"),
+            None => String::from("&str"),
+        },
         Type::Struct(name) => {
             if type_context.struct_uses_borrowed_slice(name) {
                 format!("{name}<{}>", lifetime.unwrap_or("'_"))
--- a/stage1/crates/axiomc/src/diagnostics.rs
+++ b/stage1/crates/axiomc/src/diagnostics.rs
@@ -140,7 +140,9 @@ fn repair_hint(kind: &str, _message: &str) -> Option<DiagnosticRepair> {
         ),
         "manifest" | "capability" => (
             "edit_manifest",
-            Some("Update axiom.toml with the narrowest required package, dependency, or capability change."),
+            Some(
+                "Update axiom.toml with the narrowest required package, dependency, or capability change.",
+            ),
             None,
         ),
         "import" => (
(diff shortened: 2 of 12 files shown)


athena-omt (changes requested):
The `String` -> `&str` coercion currently applies to any expression with `Type::String`, and codegen renders it as `<expr>.as_str()`. That is fine for an existing binding like `read(owned)`, but it also accepts temporaries in places where the borrowed value must outlive the statement.

For example, this source would lower successfully:

```axiom
fn make(): String {
return "owned"
}

let borrowed: &str = make()
print borrowed
```

The generated Rust shape is effectively:

```rust
let borrowed: &str = make().as_str();
```

which borrows a temporary `String` that is dropped at the end of the statement. The same issue can happen with string concatenation assigned to `&str`, since `BinaryAdd` now returns `Type::String` and then gets coerced to `&str`.

This should either be rejected during lowering/borrow checking when the `StringBorrow` operand is not a stable owner, or the coercion should only be inserted in contexts where the borrow cannot escape the expression, such as direct function-call arguments. Please add regression coverage for borrowed temporaries as well as the existing owned-variable case.

I could not run local checks because even read-only shell commands in the supplied worktree fail with `bwrap: Failed to make / slave: Permission denied`; this review is based on the PR diff from GitHub.

athena-omt (changes requested):
New blocker beyond the existing temporary-borrow issue: `&str` borrows are not integrated into the ownership/borrow tracking path.

`Type::Str` is explicitly treated as non-borrowed in `contains_borrowed_slice_type_inner`, so a binding like:

```axiom
let owned: string = "hello"
let borrowed: &str = owned
let moved: string = owned
print borrowed
```

can lower without recording `borrowed` as an active borrow of `owned`. That allows moving the `String` while the `&str` is still live, producing invalid generated Rust / violating the intended ownership model.

The PR now rejects some escaping temporary `String` borrows, but stable-owner `String -> &str` coercions also need owner tracking and move-while-borrowed protection, with regression coverage for moving the owner while an `&str` binding remains live.

athena-omt (changes requested):
The existing blocker remains: `&str` is still not treated as a borrowed type in the ownership tracking path.

In `stage1/crates/axiomc/src/hir.rs`, `contains_borrowed_slice_type_inner` still returns `false` for `Type::Str`. As a result, when lowering `let borrowed: &str = owned`, `binding_borrowed_owners_from_expr` returns an empty owner set and `borrow_kind_for_type(&Type::Str, ...)` returns `None`, so the owner `owned` never gets an active shared borrow recorded. `Expr::StringBorrow` delegates to `expr_borrowed_owners(expr, ...)`, but because the inner expression has `Type::String`, the early `contains_borrowed_slice_type(expr.ty(), ...)` guard also prevents discovering the owner there.

That means the previously discussed pattern can still lower without move-while-borrowed protection:

```axiom
let owned: string = "hello"
let borrowed: &str = owned
let moved: string = owned
print borrowed
```

Please wire `Type::Str` into the same borrowed-owner / active-borrow accounting used for slices, and add regression coverage for moving a `String` owner while an `&str` binding derived from it remains live.

athena-omt (changes requested):
Thanks for the follow-up fixes. The prior plain-owner blocker appears to be addressed for `let borrowed: &str = owned`, but there is still a new escaping-borrow hole around indexed expressions.

`is_stable_string_borrow_owner` currently treats `Expr::Index { .. }` as a stable source for `String -> &str` coercion. However, `owned_borrow_root` does not return an owner for `Expr::Index`, so a `StringBorrow` of an indexed value records no active borrowed owner. More importantly, codegen for indexing a non-Copy array element renders through the value-producing path, e.g. `axiom_array_take(...).as_str()`, which is a borrow of a temporary/moved `String`, not a stable owner.

A shape like this is therefore still accepted by the lowering rules but can generate an invalid/dangling Rust borrow:

```axiom
let values: [string] = ["hello"]
let borrowed: &str = values[0]
print borrowed
```

Please do not classify arbitrary `Index` expressions as stable `String` borrow owners unless the lowering/codegen path actually borrows from a live owner and records that owner for active-borrow accounting. The safer fix is to reject indexed `String -> &str` coercions for now, or implement explicit owner tracking/codegen for this case, with regression coverage for array/slice indexing into `String`. I could not run local checks because read-only exec fails in this review environment with `spawn /bin/sh ENOENT`, so this is based on direct inspection of `hir.rs` and `codegen.rs`.

athena-omt (changes requested):
Thanks for the follow-up fixes. The plain `let borrowed: &str = owned` owner tracking and the earlier indexed-string hole look addressed in the current lowering logic, but there is still an escaping temporary-borrow case through borrow-returning functions.\n\n`lower_call_arg_with_expected` permits temporary `String -> &str` coercions for all call arguments by passing `allow_temporary_string_borrow = true`. That is safe for ordinary sink-style calls, but not when the callee's signature returns a borrow of that parameter via `borrow_return_params`. For example, a shape like this can still be lowered:\n\n```axiom\nfn echo(value: &str): &str {\n  return value\n}\n\nfn make(): String {\n  return \"owned\"\n}\n\nlet borrowed: &str = echo(make())\nprint borrowed\n```\n\nThe generated Rust shape is effectively `let borrowed: &str = echo(make().as_str());`, so the borrowed `String` temporary is dropped at the end of the statement while the returned `&str` remains live. In the current code, `record_temporary_borrows`/`release_temporary_borrows` only models statement-local call borrows, and `expr_borrowed_owners` for the returned `Call` cannot attach an owner for `StringBorrow(make())` because `owned_borrow_root` returns `None`.\n\nPlease reject temporary `String -> &str` coercions for arguments whose parameter can be returned/escaped by the callee, or otherwise materialize/track a stable owner before allowing that call result to be bound. A regression fixture for a borrow-returning function called with `make()`/concatenation would cover this gap.

athena-omt (commented):
By inspection of the current PR head (`30c50f6`), the previously reported string-borrow blockers appear addressed.

The relevant changes now wire `Type::Str` into borrowed-type detection and shared-borrow accounting, `StringBorrow` records a stable owned root, arbitrary indexed string expressions are no longer accepted as stable borrow owners, and call arguments for borrow-returning parameters now disallow temporary `String -> &str` coercions so returned borrows cannot escape a dropped temporary.

I’m leaving this as a comment rather than an approval because I could not run local validation in this worker environment: `cargo` is not installed, and `gh` is also unavailable for independent check/merge-state verification. The assignment also reports the PR merge state as `DIRTY`, which still needs to be cleared before merge readiness.

daedalus-omt:

The replacement PR preserves the issue-closing semantics for #221 and is based on current `origin/main`, so I am closing this older conflicted branch PR in favor of the consolidated replacement.

PR #491, opened 2026-04-30
fix: address issue #220

### LINKED ISSUE #220: Language: Full numeric tower — i8..i64, u8..u64, f32, f64

daedalus-omt:
Closes #220

## Summary
- Adds support for the expanded numeric tower beyond `int` and `bool`.
- Updates syntax, HIR, MIR, diagnostics, JSON contract handling, project logic, and codegen for the new numeric forms.
- Preserves explicit typing/casting expectations for mixed-width numeric behavior.

## Scope and conformance
- Issue: Language: Full numeric tower — `i8..i64`, `u8..u64`, `f32`, `f64`.
- Primary implementation paths: `stage1/crates/axiomc/src/{syntax,hir,mir,codegen,diagnostics,json_contract,lib,project}.rs`.

## Testing

## Notes for reviewers
- This is the Daedalus-assigned implementation slice for #220.
- Please focus review on type checking, codegen mapping, and diagnostic behavior for mixed-width or unsupported numeric operations.

### DIFF
--- a/stage1/crates/axiomc/src/codegen.rs
+++ b/stage1/crates/axiomc/src/codegen.rs
@@ -2140,6 +2140,7 @@ fn expr_uses_call(expr: &Expr, name: &str) -> bool {
         Expr::BinaryAdd { lhs, rhs, .. } | Expr::BinaryCompare { lhs, rhs, .. } => {
             expr_uses_call(lhs, name) || expr_uses_call(rhs, name)
         }
+        Expr::Cast { expr, .. } => expr_uses_call(expr, name),
         Expr::Try { expr, .. }
         | Expr::Await { expr, .. }
         | Expr::FieldAccess { base: expr, .. } => expr_uses_call(expr, name),
@@ -2176,6 +2177,13 @@ struct TypeContext<'a> {
 }
 
 impl<'a> TypeContext<'a> {
+    fn empty() -> Self {
+        Self {
+            structs: HashMap::new(),
+            enums: HashMap::new(),
+        }
+    }
+
     fn new(program: &'a Program) -> Self {
         Self {
             structs: program
@@ -2210,7 +2218,12 @@ impl<'a> TypeContext<'a> {
         visiting_enums: &mut HashSet<String>,
     ) -> bool {
         match ty {
-            Type::Int | Type::Bool | Type::String | Type::Ptr(_) | Type::MutPtr(_) => false,
+            Type::Int
+            | Type::Numeric(_)
+            | Type::Bool
+            | Type::String
+            | Type::Ptr(_)
+            | Type::MutPtr(_) => false,
             Type::Slice(_) | Type::MutSlice(_) => true,
             Type::Struct(name) => {
                 if !visiting_structs.insert(name.clone()) {
@@ -2782,6 +2795,7 @@ fn render_source_marker(
 fn render_expr(expr: &Expr) -> String {
     match expr {
         Expr::Literal(LiteralValue::Int(value)) => value.to_string(),
+        Expr::Literal(LiteralValue::Numeric { raw, ty }) => format!("{raw}{}", ty.as_str()),
         Expr::Literal(LiteralValue::Bool(value)) => value.to_string(),
         Expr::Literal(LiteralValue::String(value)) => format!("String::from({value:?})"),
         Expr::VarRef { name, .. } if name == "self" => String::from("self_"),
@@ -3105,7 +3119,7 @@ fn render_expr(expr: &Expr) -> String {
             format!("{name}({rendered_args})")
         }
         Expr::BinaryAdd { lhs, rhs, ty } => match ty {
-            Type::Int => format!("{} + {}", render_expr(lhs), render_expr(rhs)),
+            Type::Int | Type::Numeric(_) => format!("{} + {}", render_expr(lhs), render_expr(rhs)),
             Type::String => format!(
                 "format!(\"{{}}{{}}\", {}, {})",
                 render_expr(lhs),
@@ -3130,6 +3144,11 @@ fn render_expr(expr: &Expr) -> String {
         Expr::BinaryCompare { op, lhs, rhs, .. } => {
             format!("{} {} {}", render_expr(lhs), op.lexeme(), render_expr(rhs))
         }
+        Expr::Cast { expr, ty } => format!(
+            "({}) as {}",
+            render_expr(expr),
+            rust_type(ty, &TypeContext::empty())
+        ),
         Expr::Try { expr, .. } => format!("({})?", render_expr(expr)),
         Expr::Await { expr, .. } => format!("axiom_await({})", render_expr(expr)),
         Expr::StructLiteral { name, fields, .. } => {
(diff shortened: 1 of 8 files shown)


athena-omt (changes requested):
`stage1/crates/axiomc/src/syntax.rs` accepts invalid unsigned suffixed literals. In `parse_numeric_literal`, the integer branch allows any suffix to pass if the number starts with `-` and the rest parses as `u128`, so `-1u8` lowers as a valid `u8` literal and codegen emits invalid Rust (`-1u8`). It also doesn’t check the literal fits the target width, so values like `300u8` are accepted by the front end and fail later in generated Rust.

This should be diagnosed during parsing/lowering for the numeric tower rather than producing invalid generated code. Please add range/sign validation per numeric suffix and tests for rejected unsigned-negative and out-of-range literals.

I could not run local checks because the read-only shell sandbox failed before executing commands, but this blocker is visible in the PR diff.

athena-omt (changes requested):
I confirmed the previous unsigned-negative / integer-width literal blocker has been addressed in `parse_numeric_literal` via `numeric_literal_fits`.

New blocker: #220 explicitly scopes `checked_*` / `wrapping_*` methods for the numeric tower, but this PR does not add any numeric method support. `hir.rs` still routes method calls only through `method_owner_name`, which returns an owner only for `Type::Struct` and `Type::Enum`; `Type::Numeric(_)` has no method dispatch path. As a result, code like `x.wrapping_add(1u8)` or `x.checked_add(1u8)` on `u8`/`i32`/etc. will still fail with “type u8 does not support method calls” rather than lowering to the required numeric operations.

Please either implement the scoped checked/wrapping numeric methods with tests, or explicitly narrow the PR/issue acceptance so this is not claimed as the full #220 fix.

athena-omt (approved):
The previous blockers look addressed in this revision:

- `parse_numeric_literal` now validates sign/range through `numeric_literal_fits`, covering unsigned-negative and out-of-range suffixed integer literals.
- Numeric method calls now get a dedicated `numeric_method_return_ty` path before struct/enum method dispatch, lower to `__axiom_numeric_*`, and render back to Rust primitive checked/wrapping calls. The added tests cover the happy path and argument-width mismatch.

I did not find a new substantive correctness blocker in the updated diff. Merge state is still `BEHIND`, but from the PR changes themselves this looks ready once queue/base-state blockers are cleared.

athena-omt (approved):
The previously raised blockers appear addressed in the current branch:

- Numeric literal parsing now models the expanded numeric tower and validates suffixed integer sign/range before lowering, so unsigned-negative and width-overflow cases are rejected in the front end rather than escaping to invalid generated Rust.
- Numeric method calls now have a dedicated lowering path for checked/wrapping operations on `Type::Numeric(_)`, enforce a matching argument type, lower to `__axiom_numeric_*`, and codegen renders them back to Rust primitive checked/wrapping calls.
- The implementation keeps the new numeric types threaded through syntax/HIR/MIR/codegen without disrupting the existing `int` path.

I did not find a new substantive correctness blocker in the inspected files. I could not run local checks because the shell command runner failed before execution with `/usr/bin/sh ENOENT`, so this approval is based on read-only source inspection. Merge state is still `BEHIND`, but from the PR changes themselves this looks ready once queue/base-state blockers are cleared.

athena-omt (changes requested):
I found a new validation hole not covered by the earlier reviews: suffixed float literals accept non-Rust float strings such as `NaNf64`/`inff32`.

`parse_numeric_literal` delegates `f32`/`f64` validation to `number.parse::<f64>().is_ok()`, so those values parse successfully, but codegen preserves the raw text and emits invalid Rust literals like `NaN` or `inf`. This lets the front end accept source that later fails in generated Rust, which is the same class of issue the integer suffix validation was meant to prevent.

Please restrict float suffixed literals to the intended numeric literal grammar/finite values, or render special values explicitly if they are part of the language, and add rejection tests.

I could not run local checks because the shell runner still fails before execution with `/bin/sh ENOENT`; this review is based on read-only source inspection.

jmcte (dismissed):

athena-omt (approved):
The latest revision appears to address the float-literal validation blocker I raised earlier.

`parse_numeric_literal` now routes `f32`/`f64` through `float_literal_is_finite_rust_number`, which rejects non-literal/special values by requiring a digit-leading Rust-like numeric form, filtering invalid characters, and requiring `parse::<f64>()` to produce a finite value. That closes the previously noted `NaNf64` / `inff32` path where the front end accepted source that codegen preserved as invalid Rust.

I also re-checked the earlier issue areas at a high level: integer suffix sign/range validation remains in place via `numeric_literal_fits`, and numeric checked/wrapping method dispatch remains dedicated to integer numeric receivers before struct/enum method lookup.

I could not run local checks because the command runner still fails before execution with `/bin/sh ENOENT`, so this approval is based on read-only source inspection. From the PR changes themselves, I do not see a remaining substantive correctness blocker.

athena-omt (changes requested):
I found one new blocker: `f32` suffixed float literals are range-checked as `f64`, so values like `1e39f32` pass the front end but will be emitted as Rust `1e39f32`, which is an overflowing `f32` literal and fails downstream. This is the same front-end/codegen validity class as the prior literal issues, but a distinct case from `NaN`/`inf`.

Please validate float suffixes against their actual target type (`f32` via `parse::<f32>().is_finite()`, `f64` via `parse::<f64>().is_finite()`) and add rejection coverage for overflowing `f32` literals.

I could not run local checks because the command runner fails before execution with `/bin/sh ENOENT`; this review is based on read-only source inspection.

athena-omt (approved):
The current branch appears to address the last blocker I raised: `f32` and `f64` suffixed literals now validate against their actual target types, with `f32` using `parse::<f32>().is_finite()` and `f64` using `parse::<f64>().is_finite()`.

I re-checked the earlier issue areas by source inspection:
- unsigned-negative and out-of-range integer suffix validation remains in `numeric_literal_fits`
- special/non-literal float strings are still rejected by the Rust-like float shape check plus finite parsing
- checked/wrapping numeric methods still lower through the dedicated `Type::Numeric(_)` path before normal struct/enum method dispatch

I could not run local checks because command execution still fails before launch with `/bin/sh ENOENT`, so this is source-inspection based. From the PR changes themselves, I do not see a remaining substantive correctness blocker.

daedalus-omt:

The replacement PR preserves the issue-closing semantics for #220 and is based on current `origin/main`, so I am closing this older conflicted branch PR in favor of the consolidated replacement.

PR #495, opened 2026-05-01
Draft trait/interface RFC for #216

### LINKED ISSUE #216: Language: Traits / interfaces with bounded generics and dyn dispatch

daedalus-omt:
Refs #216

Adds a draft RFC and disposition update for traits/interfaces with bounded generics and dyn dispatch. This is a scoping/design slice only; it does not close #216 or claim the implementation bar has landed.


athena-omt (changes requested):
Blocking on the issue lifecycle, not on the RFC text itself.

Issue #216 was explicitly reopened with this closure bar: it should remain open unless a PR lands real implementation work or clear already-shipped evidence for the requested feature, and disposition-only documentation is not enough for closure. This PR adds a draft RFC and updates the disposition docs, but the PR body says `Closes #216`, so merging it would close the governing issue without meeting that stated bar.

Please change the PR linkage so this references #216 without auto-closing it, or pair this with the owner-approved split/implementation issue transition that satisfies the closure condition. The RFC scoping itself looks directionally consistent with the existing disposition text.

Summary: request changes until the PR no longer auto-closes #216 as a docs-only RFC scoping change.

jmcte:
Addressed the requested issue-lifecycle blocker: the PR now uses `Refs #216` and is titled as a draft RFC/design slice, so merging it will not auto-close #216 or claim that the implementation closure bar has landed.

athena-omt (approved):
Approved. I rechecked the prior lifecycle blocker rather than duplicating it: the PR body now uses `Refs #216`, and the new RFC explicitly states that it scopes follow-up implementation work and deliberately does not close #216. The disposition table also continues to keep #216 open while documenting the proposed static-bounds-before-`dyn` split.

I did not find a new correctness or scope blocker in the docs diff. No executable checks are required for this RFC-only/design slice.

Summary: approve; the previous changes-requested condition has been satisfied and this is ready as a draft RFC update.

athena-omt (approved):
Approved. I rechecked the live PR body and diff rather than repeating the earlier lifecycle review: the PR now references #216 without auto-closing it, the RFC explicitly says it does not close #216, and the disposition table still keeps #216 open while documenting the proposed static-bounds-before-`dyn` split.

The staged scope is consistent with the current stage1 architecture: parser/static bounds and monomorphized dispatch first, with `dyn`, coherence, blanket impls, and stdlib trait stabilization deferred behind explicit follow-up design/implementation work. CI is green, and no executable checks beyond the existing PR checks are needed for this docs/RFC slice.

Summary: approve; ready to merge once unrelated queue/branch protection blockers clear.

pheidon (approved):
Approved. I did a second-pass read of the RFC/disposition updates and verified the governance link locally: the RFC explicitly governs #216 while saying it does not close #216, and the disposition table keeps #216 open pending implementation or owner-approved split. This is the right non-closing RFC shape.

PR #499, opened 2026-05-01
Add duplicate export diagnostics

### LINKED ISSUE #342: Modules: duplicate export and ambiguous symbol diagnostics

Make public API collisions explicit when imports flatten symbols into the current namespace. Acceptance: duplicate pub fn/type/struct and imported/exported conflicts fail with precise spans.

Working rules:

jmcte:
## Summary
- reject local cross-kind symbol collisions before module flattening can create ambiguous public surfaces
- reject imported public/package symbols that collide with existing visible names across function, const, alias, struct, and enum namespaces
- add focused stage1 regressions for local and imported export collisions

Closes #342

## Validation
- cargo test --manifest-path stage1/Cargo.toml -p axiomc export_symbol_collision --quiet
- cargo fmt --manifest-path stage1/Cargo.toml
- cargo test --manifest-path stage1/Cargo.toml -p axiomc import --quiet
- git diff --check
- make stage1-smoke


pheidon (approved):
Reviewed the diff for duplicate export diagnostics. The change adds explicit cross-namespace collision checks both within a module and when flattening imported public/package exports, while preserving the existing same-namespace duplicate handling. The new tests cover local struct/function collisions and imported type/function collisions, and current required checks are green. I did not find a blocking issue.

PR #503, opened 2026-05-02
fix: address issue #422

### LINKED ISSUE #422: Bench: Go/Rust comparison gate

Add simple equivalent Go/Rust/Axiom workload comparison. Acceptance: reports build time, run time, binary size, diagnostics quality, and capability manifest coverage without blocking PRs initially.

Working rules:

apollo-omt:
Closes #422

Implements the Apollo-assigned fix for: Bench: Go/Rust comparison gate


athena-omt (dismissed):
Approved. The PR adds the stage1 Go/Rust/Axiom comparison as an explicitly advisory report, wires it into fast CI with `continue-on-error`, uploads the JSON artifact when present, and documents the non-blocking policy plus future `--enforce` path. The benchmark script keeps the measured workloads scoped, reports medians/samples/binary sizes/diagnostic/capability coverage, and only returns nonzero when `--enforce` is used and advisory limits fail. I did not find a substantive correctness or merge-readiness blocker in the changed files. The branch is currently behind `main`, but that is queue/branch hygiene rather than a code-review blocker.

athena-omt (changes requested):
Requesting changes. I found a merge-readiness blocker that is not covered by the prior approval: in `.github/workflows/pr-fast-ci.yml`, the `ci-gate` step uses `if [["$status" == "success" || "$status" == "skipped"]]; then`. Bash requires spaces after `[[` and before `]]`, so this parses as an invalid command/condition and will cause the CI Gate job to fail when it runs. Please change it to the normal spaced form, e.g. `if [[ "$status" == "success" || "$status" == "skipped" ]]; then`.

Secondary issue: `docs/performance-benchmarks.md` says the extended validation gate compares against `stage1/benchmarks/baselines/stage1-build-median.json`, but `check-stage1-benchmarks.py` only defines `load_regression_baseline()` / `compare_regression_baseline()` and never calls them or includes their warnings in the report. Either wire that comparison into the JSON/advisory output or remove the claim from the docs.

Compact summary: fix the CI Gate shell syntax before merge; align the baseline-comparison docs with the actual script behavior.

athena-omt (approved):
Approved. I re-reviewed the PR branch diff and specifically checked the prior blockers: the CI Gate shell condition now uses valid spaced `[[ ... ]]` syntax, and the benchmark script now actually loads and compares `stage1/benchmarks/baselines/stage1-build-median.json` via `compare_regression_baseline()` and reports the non-blocking warnings in the JSON output. I also ran `python3 -m py_compile scripts/ci/check-stage1-benchmarks.py`, `python3 scripts/ci/check-stage1-benchmarks.py --help`, and `git diff --check origin/main...HEAD`; those passed.

The added workflow remains advisory via `continue-on-error`, uploads the report artifact if present, and the `--enforce` path is opt-in. I do not see a remaining substantive code-review blocker in the changed files. The branch is still behind `main`, but that is merge-queue/branch hygiene rather than a code correctness objection.

jmcte (approved):

PR #506, opened 2026-05-02
fix: address issue #365

### LINKED ISSUE #365: Conformance: Go/Rust-style comparison fixtures

Create small programs representing strict typing, package imports, owned resources, and predictable diagnostics. Acceptance: each fixture has expected output or expected diagnostic and is marked Axiom-owned.

Working rules:

apollo-omt:
Closes #365

Implements the Apollo-assigned fix for: Conformance: Go/Rust-style comparison fixtures

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[env]
+RUST_MIN_STACK = "8388608"
--- a/stage1/conformance/README.md
+++ b/stage1/conformance/README.md
@@ -25,6 +25,9 @@ Current executable fixtures cover:
   control flow.
 - `collection_operations`: standard collection helpers over arrays and
   borrowed slices.
+- `comparison_package_resources`: Axiom-owned Go/Rust-style comparison fixture
+  for explicit package imports, strict struct typing, owned resource transfer,
+  borrowed slices, and machine-checkable result output.
 - `package_local_modules`: nested package-local module imports that execute
   successfully.
 - `package_visibility`: `pub(pkg)` items imported across sibling modules within
@@ -43,6 +46,10 @@ Current compile-fail fixtures cover:
 
 - `mutable_borrow_while_shared_live`: ownership diagnostics for conflicting
   mutable and shared borrows.
+- `comparison_owned_resource_move`: Axiom-owned Go/Rust-style comparison
+  diagnostic for an owned resource consumed by a function and then reused.
+- `comparison_strict_type_mismatch`: Axiom-owned Go/Rust-style comparison
+  diagnostic for strict struct field typing.
 - `ownership_use_after_move`: ownership diagnostics for reading a moved value.
 - `panic_rejects_unreachable_statement`: control diagnostics for statements
   that appear after `panic(...)` in the same block.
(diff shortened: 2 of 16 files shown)


athena-omt (commented):
I reviewed the conformance fixture diff for issue #365. The new pass/fail fixtures are narrow, issue-backed, include machine-checkable expected output/diagnostics, and the corpus count updates line up with the added 2 fail + 1 pass cases. GitHub checks are all passing.

I’m leaving this as a comment rather than an approval because the PR currently reports `mergeStateStatus: DIRTY`. Once the branch is rebased/updated and the conflict state clears, I don’t see a substantive code/test blocker in this slice.

athena-omt (commented):
I did a second-pass read of the added conformance fixtures and expected artifacts.

The new cases look internally consistent:
- `comparison_owned_resource_move` exercises reuse after consuming an owned struct value, with the expected `use_after_move` diagnostic pointing at `handle` on the final `print handle.name`.
- `comparison_strict_type_mismatch` cleanly covers strict struct field typing by assigning a string to an `int` field, and the expected diagnostic location/message match that source shape.
- `comparison_package_resources` covers package import, strict resource struct typing, owned array transfer, borrowed slicing, `Result` matching, and its expected output matches the execution path (`go-rust comparison`, slice length `2`, first borrowed item `20`).
- The conformance workspace and README enumerate the added fixtures, so the corpus wiring appears deliberate rather than orphaned.

I’m leaving this as a comment, not an approval, because the PR still reports `mergeStateStatus: DIRTY`, which is the same merge-readiness blocker already called out. I don’t see a new code/test correctness issue in this slice.

athena-omt (dismissed):
I reviewed the current PR branch fixtures directly in the read-only worktree.

The added comparison conformance cases are still scoped to issue #365 and appear correctly wired:
- the new fail/pass packages are present in the conformance workspace manifest;
- the strict type mismatch and owned-resource-move fail cases have diagnostics matching the source shapes and locations;
- the package resources pass case exercises import/resource/slice/Result behavior and its expected output matches the visible execution path.

I do not see a substantive correctness, architecture, or test-adequacy blocker in this slice. The branch is currently `BEHIND`, so it still needs to be updated with `main` before merge, but I’m treating that as merge readiness rather than a reason to block this review.

athena-omt (commented):
I confirm the existing review posture: the added comparison conformance fixtures still look correctly scoped and wired from read-only inspection of the PR worktree.

One small additional note from this pass: `stage1/conformance/README.md` appears to list `comparison_owned_resource_move` and `comparison_predictable_diagnostic` twice in the compile-fail fixture section. That looks like harmless documentation duplication, not a substantive correctness blocker.

I’m leaving this as a comment because the PR is still reported as `BEHIND`; update the branch before merge.

athena-omt (commented):
I could not complete a substantive review under the assignment constraints because the required read-only worktree path does not exist:

`/home/pheidon/src/_worktrees/apollo/omt-global/axiom/issue-365`

I checked for the specified path and nearby worktree roots, and only Athena/Daedalus axiom worktrees are present. Because the instructions require working only inside that exact PR branch worktree and not substituting another checkout, I’m leaving this as a comment rather than approving or requesting changes.

Please recreate/sync the required PR worktree for `apollo/issue-365`, then rerun review.

jmcte (approved):

PR #525, opened 2026-05-06
Consolidate mutable refs, numeric tower, and string ownership (#218, #220, #221)

### LINKED ISSUE #218: Language: Mutable references &mut T with exclusive-xor-shared aliasing

### LINKED ISSUE #220: Language: Full numeric tower — i8..i64, u8..u64, f32, f64

### LINKED ISSUE #221: Language: Distinguish owned String from borrowed &str

daedalus-omt:
## Summary
- Combines the compatible type-system branches on current main.
- Keeps the mutable reference, numeric tower, and string ownership fixes together to avoid repeated HIR conflicts.

## Supersedes
- Supersedes PR #493
- Supersedes PR #491
- Supersedes PR #490

## Issue closure
- Closes #218
- Closes #220
- Closes #221

## Validation
- `git diff --check`
- `cargo fmt --manifest-path stage1/Cargo.toml --check`

Not run locally: `bash scripts/ci/run-fast-checks.sh` / `cargo check --manifest-path stage1/Cargo.toml` because this execution image lacks `make` and `cc`; GitHub CI should provide the full toolchain.


jmcte (dismissed):

jmcte (dismissed):

athena-omt (changes requested):
I found a substantive type-system blocker.

The PR adds `Type::Numeric(_)` and `Type::Str`, but `hir::Type` has a hand-written `PartialEq` that never treats either variant as equal to itself. That means downstream `type_assignable_to(..., ...)` falls through to `actual == expected` and rejects same-type numeric/string-borrow assignments and comparisons.

Evidence:
- `stage1/crates/axiomc/src/hir.rs:216-222` defines `Numeric` and `Str`
- `stage1/crates/axiomc/src/hir.rs:240-268` only returns true for `Error`, `Int`, `Bool`, `String`, aggregates, and containers — not `Numeric(lhs) == Numeric(rhs)` or `Str == Str`
- `stage1/crates/axiomc/src/hir.rs:272-304` relies on that equality fallback for assignability

Validation note: `git diff --check` passed. `cargo` and `gh` are not installed in this worker image, so I could not run Rust checks or inspect live GitHub status from here.

athena-omt (changes requested):
I confirm the existing type-system blocker and would keep this as changes requested.

`hir::Type` now includes `Type::Numeric(_)` and `Type::Str`, but its hand-written `PartialEq` still does not compare either variant as equal to itself. Because `type_assignable_to` falls back to `actual == expected` after the structural cases, valid same-type numeric and borrowed-string assignments can be rejected.

Additional validation from this pass:
- `git diff --check origin/main...HEAD` passed.
- `cargo` is not installed in this worker image, so I could not run `cargo check`/tests here.

This should be fixed before merge by adding equality coverage for `Numeric(lhs) == Numeric(rhs)` and `Str == Str`, then adding/confirming focused coverage for same-type numeric and `&str` assignability.

athena-omt (changes requested):
The previous Athena blocker is fixed: `Type::Numeric(_)` and `Type::Str` now participate correctly in `hir::Type` equality. CI is green and the PR body preserves `Closes #218`, `Closes #220`, and `Closes #221`.

I still found one ownership blocker before this can merge: a borrowed `&str` returned from a local `String` can currently pass HIR lowering. In `expr_borrow_origin`, `Expr::StringBorrow { expr, .. }` delegates to `expr_borrow_origin(expr, ...)`; for the underlying `String` owner that returns `None` because `String` is not a borrowed type. The return validation then accepts `None`, so a function like `fn pick(input: &str): &str { let owned: String = "local"; return owned }` is accepted by the stage1 ownership checker and only fails later as generated Rust borrowing a local. Please make `StringBorrow` over a local/owner report `BorrowOrigin::Local` (while preserving parameter-derived origins) and add a regression test for returning a local `String` as `&str`.

jmcte:
athena-omt, I pushed fixes for the outstanding blocker and the merge conflict cleanup.

What changed:
- `StringBorrow` now preserves parameter-derived borrow origins and falls back to `BorrowOrigin::Local` for local/owned string borrows, so returning a local `String` as `&str` is rejected during HIR ownership checking.
- Added regression coverage for returning a local `string` as `&str`.
- Merged current `main` into this branch and resolved the `borrowck` split conflict.
- Fixed the follow-up CI regression where mutable slice parameters were rendered as `mut values: &mut [...]` after the merge.

Validation:
- GitHub checks are green on `e4dd8ae`.
- Local focused checks passed: `cargo fmt --manifest-path stage1/Cargo.toml --check`, `cargo check --manifest-path stage1/Cargo.toml -p axiomc`, `cargo test --manifest-path stage1/Cargo.toml check_project_rejects_str_return_from_local_string`, `cargo test --manifest-path stage1/Cargo.toml parser_lowers_explicit_lifetime_slice_signatures`, `cargo test --manifest-path stage1/Cargo.toml parser_lowers_mutable_slice_signatures`, `git diff --check`, and `make stage1-smoke`.

PR is now `MERGEABLE`; the remaining block is the stale `CHANGES_REQUESTED` review state. Please re-review when you can.

athena-omt (approved):
Re-reviewed the latest head `e4dd8ae` after John’s follow-up comment. The prior ownership blocker is fixed: `Expr::StringBorrow` now preserves derived borrow origins and falls back to `BorrowOrigin::Local`, so returning a local `string` as `&str` is rejected during HIR ownership checking. Regression coverage is present in `check_project_rejects_str_return_from_local_string`.

I also confirmed the merge-conflict cleanup around the borrow-check split has no conflict markers, `Type::Str` is included in borrowed-type detection, and the mutable-slice rendering regression is covered by `parser_lowers_mutable_slice_signatures` / explicit lifetime signature assertions so parameters render as `values: &mut [...]` rather than `mut values: &mut [...]`.

Validation I could run from the Athena checkout: `git fetch origin main pull/525/head`, inspected the PR diff at `origin/main...HEAD`, `git diff --check origin/main...HEAD` passed, and live GitHub check-runs for `e4dd8ae` are green. This Athena worker image still lacks `cargo`, `rustc`, `make`, and `gh`, so I could not rerun the local Rust commands here; I verified their targeted coverage/code paths by inspection and relied on the live green GitHub check-runs for CI truth.

pheidon (approved):
Pheidon sweep review: approved on latest head 3eaf3d7. I reviewed this after Athena’s latest approval and John’s follow-up fixes. The prior blockers are addressed: hir::Type equality now accounts for the new numeric/string variants, and StringBorrow preserves parameter-derived borrow origins with the fallback root path for owned strings.

I inspected the ownership/type-system/codegen/syntax changes and the added ownership failure fixtures. The consolidation is large, but the changes are internally consistent: mutable refs, numeric tower handling, borrowed str/string ownership, fixture coverage, and docs all line up.

Validation run locally:
- git diff --check origin/main...HEAD
- cargo test --manifest-path stage1/Cargo.toml -p axiomc

Result: 397 lib tests + 18 main tests passed, doc-tests passed. No blocker found.

Note: GitHub currently reports the branch as BEHIND, so it may still need an update/merge-main before it is mergeable, but I do not see a review blocker in the PR content.

PR #558, opened 2026-05-10
feat(stage1): add URL encoding helpers

### LINKED ISSUE #405: Stdlib: URL and path encoding helpers

Provide small helpers for safe URL query/path construction. Acceptance: HTTP client/service fixtures avoid hand-built escaped strings.

Working rules:

daedalus-omt:
## Summary
- Adds stage1 stdlib percent-encoding helpers for URL components, query pairs, and path-segment joins.
- Extends the encoding example fixture to cover query/path helpers.
- Updates the HTTP client fixture to construct encoded path/query pieces with helpers instead of hand-built escaped strings.

Closes #405

## Checks
- git diff --check
- cargo test -p axiomc stdlib_encoding -- --nocapture (running on Daedalus with isolated CARGO_TARGET_DIR; initial attempts blocked on a shared cargo artifact lock)

## Merge Automation
Auto-merge requested when GitHub permits it.

### DIFF
--- a/docs/stage1-encoding-helpers.md
+++ b/docs/stage1-encoding-helpers.md
@@ -8,6 +8,10 @@
   percent escapes or invalid UTF-8.
 - `path_segment_encode(value)` uses the same percent-encoding contract for one
   path segment, including escaping `/`.
+- `query_pair_encode(name, value)` returns `name=value` with both sides encoded
+  as URL components.
+- `path_join_segment(base, segment)` appends one encoded segment to a base path,
+  inserting exactly one `/` separator when the base is non-empty.
 
 These helpers are pure string utilities and do not grant network or filesystem
 capabilities.
--- a/stage1/crates/axiomc/src/codegen.rs
+++ b/stage1/crates/axiomc/src/codegen.rs
@@ -895,6 +895,23 @@ fn axiom_percent_encode(value: String) -> String {
 }
 
 #[allow(dead_code)]
+fn axiom_query_pair_encode(name: String, value: String) -> String {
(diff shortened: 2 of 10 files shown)


athena-omt (changes requested):
Thanks for the focused stdlib expansion. I found one merge-blocking issue:

- `stage1/examples/stdlib_http/src/main.ax:6` and `stage1/examples/stdlib_http/src/main_test.ax:6` call `query_pair("q", "hello world/one")`, but the new exported helper is `query_pair_encode(name, value)`. That leaves the updated HTTP fixture using an undefined function. The same files also import `std/encoding.ax` twice.

Please update those calls to `query_pair_encode(...)` and drop the duplicate import. I verified `git diff --check` is clean; full cargo validation was not completed before the worker command timeout, but the source-level compile blocker is direct.

athena-omt (approved):
Approved. I re-reviewed the current PR head (`3e33337`) in the Athena worker worktree. The previously requested fix is present: `stage1/examples/stdlib_http/src/main.ax` and `stage1/examples/stdlib_http/src/main_test.ax` now call `query_pair_encode(...)`, and the duplicate `std/encoding.ax` import is gone.

`git diff --check origin/main...HEAD` is clean. The new helpers are wired through stdlib, HIR lowering, and codegen consistently, and the updated fixtures cover the new query/path helper behavior. I do not see a remaining source-level merge blocker.

Note: I attempted the targeted cargo validation, but the worker command timed out before returning test output, so this approval is based on source review plus diff hygiene rather than a completed local cargo run.

jmcte (approved):

PR #570, opened 2026-05-12
docs: define write capability boundary RFC

### LINKED ISSUE #391: Capabilities: write-capability boundary RFC

Design package-root-scoped write/copy/mkdir rules without implementing broad filesystem mutation. Acceptance: RFC/doc plus failing fixtures define allowed and denied behavior.

Working rules:

apollo-omt:
## Summary
- Add RFC 0002 defining the package-root-scoped `fs:write` boundary for write/copy/mkdir capability behavior.
- Register a failing stage1 conformance fixture proving stdlib write helpers require `[capabilities].fs:write = true`.
- Document allowed and denied write/copy behavior without broadening host filesystem mutation.

Closes #391.

## Checks
- `make stage1-conformance`

## Merge Automation
Auto-merge requested with squash after PR creation.

### DIFF
--- a/docs/rfcs/0002-write-capability-boundary.md
+++ b/docs/rfcs/0002-write-capability-boundary.md
@@ -1,0 +1,149 @@
+# RFC 0002: Write-Capability Boundary
+
+Status: Draft
+
+Governing issue: #391
+
+## Summary
+
+Axiom should expose filesystem mutation only through an explicit, package-root-scoped `fs:write` capability. The initial write surface is intentionally narrow: create/write/append/replace file content and create directories inside the package capability root. Copy semantics are specified as a future helper over the same boundary, but this RFC does not broaden host filesystem mutation or add ambient delete/rename authority.
+
+## Motivation
+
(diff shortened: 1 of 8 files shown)


jmcte (approved):

athena-omt (changes requested):
Thanks — I verified the new compile-fail fixture and ran `make stage1-conformance` successfully: 64/64 passed.

I do need one doc-contract fix before this is merge-ready. RFC 0002 states that the initial `fs:write` surface has “no ... delete ... authority” and lists allowed operations without delete helpers, but the current `std/fs.ax` write surface already exposes `remove_file` and `remove_dir`, and `capability_for_call` gates both on `fs:write`. The RFC later acknowledges possible future removal of broad delete helpers, so the document is internally aware of that reality, but the normative sections currently define a boundary that does not match the implementation.

Please either include the existing non-recursive delete helpers in the RFC’s `fs:write` boundary with the same root-scoping/denial rules, or explicitly mark them as legacy/deprecated and clarify that this RFC’s “initial surface” is the target contract rather than the current stage1 API. Without that clarification, schedulers/reviewers could read the RFC and incorrectly assume `fs:write` does not authorize deletion today.

PR #578, opened 2026-05-12
Add process command capability allowlists

### LINKED ISSUE #393: Capabilities: process command allowlists

Replace broad process_status(command) strings with manifest-scoped command allowlists. Acceptance: unlisted commands fail at check/build time; no shell interpolation by default.

Working rules:

daedalus-omt:
## Summary
- Add manifest-scoped process command allowlists via `[capabilities] process = [\"...\"]`.
- Surface allowlisted process commands in capability descriptors/contracts.
- Reject `process_status` calls at project check time when the command is not listed, and reject dynamic process command expressions when an allowlist is configured.
- Preserve current generated runtime behavior of `std::process::Command::new(program)`, avoiding shell interpolation by default.

Closes #393

## Checks
- `cargo test -p axiomc process_command -- --nocapture` attempted on Daedalus; blocked by pre-existing/unrelated compile failures in current branch/base: duplicate static-global tests in `crates/axiomc/src/lib.rs`, missing `create_project` test helper references in `crates/axiomc/src/main.rs`, and missing `http` field in a `TestTarget` initializer in `crates/axiomc/src/project.rs`.
- `cargo test -p axiomc --lib process_command -- --nocapture` attempted on Daedalus; hit the same unrelated compile blockers.
- `cargo fmt --check` attempted on Daedalus; blocked by broad pre-existing formatting drift outside this slice, including `axiom-lsp`, `axiomc/src/main.rs`, and other files not touched for this issue.

## Merge Automation
Auto-merge not enabled yet because the local validation commands cannot compile due to unrelated current-tree blockers described above.

### DIFF
--- a/stage1/crates/axiomc/src/hir.rs
+++ b/stage1/crates/axiomc/src/hir.rs
@@ -7574,6 +7574,7 @@ fn lower_expr_with_expected_inner(
                     .with_span(*line, *column));
                 }
                 let lowered = lower_expr_with_expected(&args[0], Some(&Type::String), env, ctx)?;
+                validate_process_command_allowlist_hir(ctx.capabilities, &lowered, *line, *column)?;
                 if lowered.ty() != &Type::String {
                     return Err(Diagnostic::new(
                         "type",
@@ -9734,6 +9735,37 @@ fn validate_ffi_type(ty: &Type, line: usize, column: usize) -> Result<(), Diagno
         _ => Err(Diagnostic::new(
             "type",
             "FFI signatures only support int, bool, string, ptr<T>, and mutptr<T> in stage1",
+        )
+        .with_span(line, column)),
+    }
+}
+
+
+fn validate_process_command_allowlist_hir(
+    capabilities: &CapabilityConfig,
+    command: &Expr,
+    line: usize,
+    column: usize,
+) -> Result<(), Diagnostic> {
+    if capabilities.process_commands.is_empty() {
+        return Ok(());
+    }
+    match command {
+        Expr::Literal {
+            value: LiteralValue::String(value),
+            ..
+        } if capabilities.process_commands.iter().any(|allowed| allowed == value) => Ok(()),
+        Expr::Literal {
+            value: LiteralValue::String(value),
+            ..
+        } => Err(Diagnostic::new(
+            "capability",
+            format!("call to \"process_status\" requires [capabilities].process to include {value:?}"),
+        )
+        .with_span(line, column)),
+        _ => Err(Diagnostic::new(
+            "capability",
+            "call to \"process_status\" requires a string literal listed in [capabilities].process",
         )
         .with_span(line, column)),
     }
--- a/stage1/crates/axiomc/src/lib.rs
+++ b/stage1/crates/axiomc/src/lib.rs
@@ -3838,6 +3838,42 @@ process = ["/bin/true"]
     }
 
     #[test]
+    fn load_manifest_rejects_empty_process_allowlist() {
+        let dir = tempdir().expect("tempdir");
+        let project = dir.path().join("process-empty-allowlist");
+        create_project(&project, Some("process-empty-allowlist-app")).expect("create project");
+        fs::write(
+            project.join("axiom.toml"),
+            r#"[package]
+name = "process-empty-allowlist-app"
+version = "0.1.0"
+
+[build]
+entry = "src/main.ax"
+out_dir = "dist"
+
+[capabilities]
+process = []
+"#,
+        )
+        .expect("write manifest");
+        fs::write(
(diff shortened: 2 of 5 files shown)


athena-omt (changes requested):
Blocking finding:

- `stage1/crates/axiomc/src/project.rs:2752` only applies the new allowlist check when the syntax-level call name is exactly `process_status`.
- The stdlib wrapper in `stage1/crates/axiomc/src/stdlib.rs:156-157` exposes `run_status(command)` and forwards to `process_status(command)`.
- That wrapper is validated under the synthetic stdlib manifest, whose process capability is unrestricted (`process_commands: Vec::new()` in `project.rs:1137`), while the later HIR capability check in `hir.rs:7561-7589` only checks that process is enabled and does not enforce the command allowlist.

So a package with:

```toml
[capabilities]
process = ["/bin/true"]
```

can plausibly call:

```axiom
import "std/process.ax"
print run_status("/bin/false")
```

and bypass the manifest-scoped command allowlist. That breaks the core security property this PR is adding.

I’d request changes to enforce the allowlist after import/stdlib lowering as well, or otherwise make stdlib process wrappers preserve callsite literal/allowlist validation. Please add a regression test covering `std/process.ax` with an allowlist that excludes the invoked command.

I did not run the Rust test suite because the PR already documents unrelated current-tree compile blockers; this review is based on local diff/source inspection in the routed Athena worktree.

jmcte (dismissed):

athena-omt (changes requested):
`[capabilities] process = []` currently normalizes to `process: true` with an empty `process_commands` vector in `manifest.rs:692-694`. Both allowlist validators then treat an empty vector as “no allowlist configured” and return `Ok(())` (`project.rs:2831-2832`, `hir.rs:9750-9751`). The schema also permits an empty array because the array branch has no `minItems`.

That means a manifest author can write what looks like an empty allowlist / deny-all policy:

```toml
[capabilities]
process = []
```

…but the compiler will enable unrestricted `process_status(...)` instead. For a security capability allowlist, that fail-open behavior is surprising and dangerous.

Please either reject empty array-form process allowlists at manifest/schema validation time, or represent “array form configured but empty” distinctly from legacy `process = true` so it denies all process commands. Add a regression test for `process = []` with `process_status("/bin/true")` to lock the intended behavior.

I did not complete Rust test execution here; `cargo test` from `stage1` timed out in the worker tool, so this review is based on direct diff/source inspection.

pheidon (dismissed):

athena-omt (approved):
Re-reviewed the current head after Pheidon's follow-up fix.

The remaining requested-change blocker is addressed: array-form `[capabilities] process = []` now fails closed at manifest normalization instead of enabling unrestricted process execution, and the manifest JSON schema now requires at least one item for process allowlist arrays.

Additional validation/evidence:
- New regression coverage rejects an empty process allowlist before `process_status("/bin/true")` can compile.
- The previous stdlib wrapper bypass blocker remains addressed: `run_status(...)` is covered by the syntax-level allowlist validation, so `std/process.ax` cannot use a command excluded by the package manifest.
- `git diff --check` passed locally on the PR head.
- `cargo check --manifest-path stage1/Cargo.toml -p axiomc --quiet` passed locally with existing warnings.
- Focused `cargo test` execution is still blocked by unrelated pre-existing axiomc test-target compile errors already observed on this branch, including duplicate test definitions and a missing `http` field in a test initializer.

I do not see a remaining PR-content blocker for the process command allowlist change.

pheidon (approved):
Pheidon re-approval after follow-up fix 14f4b05. Athena has re-reviewed and approved; CI Gate and Fast Checks are green.

athena-omt (changes requested):
I found a separate blocker beyond the existing allowlist-bypass / empty-list reviews:

`stage1/crates/axiomc/src/project.rs:2752` special-cases every syntax-level call named `run_status`, not just the stdlib wrapper. Because this check runs before name resolution, a user-defined function named `run_status` is incorrectly treated as a process capability wrapper whenever `[capabilities].process` is an allowlist.

Example that should be harmless:

```axiom
fn run_status(command: string): int {
  return 0
}

print run_status("not-allowlisted")
```

with:

```toml
[capabilities]
process = ["/bin/true"]
```

This would be rejected even though it never calls `process_status`. The HIR-level `process_status` validation is the right layer for resolved/lowered intrinsic enforcement; the syntax-level `run_status` name check is over-broad.

Please avoid enforcing process allowlists based solely on unresolved user-call names, and add a regression test proving a local/user `run_status` function is not constrained by process command allowlists.

Compact summary: not merge-ready; the PR still has security semantics under review, and this branch also introduces a false-positive capability rejection for ordinary user functions named `run_status`.

PR #587, opened 2026-05-13
HIR: audit capability attribution before MIR

### LINKED ISSUE #353: HIR: capability attribution audit pass

Report intrinsic and stdlib capability requirements from HIR before MIR/codegen. Acceptance: capability attribution tests run without native build execution.

Working rules:

daedalus-omt:
## Summary

Reports capability requirements from the lowered HIR path before MIR/codegen. The capability SBOM now lowers the flattened module graph with an audit-only capability context, then walks HIR intrinsic calls with preserved call-site spans so agents can inspect required capability surfaces without native build execution.

## Changes

- Add HIR call-site span preservation for intrinsic calls.
- Add an audit-only HIR capability attribution pass for `axiomc caps --manifest-json` / SBOM output.
- Keep capability-denial behavior separate from attribution so missing grants do not hide required capability usage.

## Testing

- `git diff --check`
- `cd stage1 && cargo fmt --all --check`
- `cd stage1 && cargo test -p axiomc capability_sbom --lib`

## Merge Automation

Auto-merge is enabled with squash; current merge state is blocked pending required checks/review gates.

Fixes OMT-Global/axiom#353

### DIFF
--- a/stage1/crates/axiomc/src/hir.rs
+++ b/stage1/crates/axiomc/src/hir.rs
@@ -172,6 +172,7 @@ pub enum Expr {
         name: String,
         args: Vec<Expr>,
         ty: Type,
+        span: SourceSpan,
     },
     BinaryAdd {
         lhs: Box<Expr>,
@@ -6439,6 +6440,10 @@ fn lower_expr_with_expected_inner(
                 }
                 move_lowered_value(&lowered, env)?;
                 return Ok(Expr::Call {
+                    span: SourceSpan {
+                        line: *line,
+                        column: *column,
+                    },
                     name: name.clone(),
                     args: with_assert_location(vec![lowered], *line, *column),
                     ty: Type::Int,
@@ -6474,6 +6479,10 @@ fn lower_expr_with_expected_inner(
                 move_lowered_value(&label, env)?;
                 move_lowered_value(&holds, env)?;
                 return Ok(Expr::Call {
+                    span: SourceSpan {
+                        line: *line,
+                        column: *column,
+                    },
                     name: name.clone(),
                     args: with_assert_location(vec![label, holds], *line, *column),
                     ty: Type::Int,
@@ -6507,6 +6516,10 @@ fn lower_expr_with_expected_inner(
                 move_lowered_value(&actual, env)?;
                 move_lowered_value(&expected, env)?;
                 return Ok(Expr::Call {
+                    span: SourceSpan {
+                        line: *line,
+                        column: *column,
+                    },
                     name: name.clone(),
                     args: with_assert_location(vec![label, actual, expected], *line, *column),
                     ty: Type::Int,
@@ -6545,6 +6558,10 @@ fn lower_expr_with_expected_inner(
                 move_lowered_value(&haystack, env)?;
                 move_lowered_value(&needle, env)?;
                 return Ok(Expr::Call {
+                    span: SourceSpan {
+                        line: *line,
+                        column: *column,
+                    },
                     name: name.clone(),
                     args: with_assert_location(vec![haystack, needle], *line, *column),
                     ty: Type::Int,
@@ -6604,6 +6621,10 @@ fn lower_expr_with_expected_inner(
                 lowered_args.push(lhs);
                 lowered_args.push(rhs);
                 return Ok(Expr::Call {
+                    span: SourceSpan {
+                        line: *line,
+                        column: *column,
+                    },
                     name: name.clone(),
                     args: with_assert_location(lowered_args, *line, *column),
                     ty: Type::Int,
@@ -6629,6 +6650,10 @@ fn lower_expr_with_expected_inner(
                     .with_span(args[0].line(), args[0].column()));
                 }
                 return Ok(Expr::Call {
+                    span: SourceSpan {
+                        line: *line,
+                        column: *column,
+                    },
                     name: name.clone(),
                     args: vec![lowered],
                     ty: Type::Int,
@@ -6657,6 +6682,10 @@ fn lower_expr_with_expected_inner(
                 }
                 move_lowered_value(&lowered, env)?;
                 return Ok(Expr::Call {
+                    span: SourceSpan {
+                        line: *line,
+                        column: *column,
+                    },
                     name: name.clone(),
                     args: vec![lowered],
                     ty: Type::Int,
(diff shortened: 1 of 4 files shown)


athena-omt (changes requested):
Thanks for moving attribution later in the pipeline. I found one blocker: `capability_sbom()` now calls `hir::lower_with_capabilities(&flattened, &context.manifest.capabilities)` before collecting intrinsic use. That lowering path enforces capability grants through `require_capability`, so a package that calls e.g. `clock_now_ms()` without `[capabilities].clock = true` will error during SBOM generation instead of producing the SBOM with the intrinsic attributed to the denied capability surface. The PR summary says the existing JSON shape for requested/denied attribution is preserved, but this changes denied-capability cases from reportable to fatal.

Please make the SBOM audit lowering path collect post-flattening/HIR attribution without enforcing the package's capability grants, or otherwise add a non-fatal audit mode that still records denied intrinsic requirements.

athena-omt (changes requested):
Thanks for adding the non-fatal audit lowering path; that addresses the prior denied-capability blocker. I found one remaining correctness issue in the new HIR collector: `collect_hir_expr_intrinsic_use` records each intrinsic using only the enclosing statement span passed down from `collect_hir_stmt_intrinsic_use` (`project.rs` around lines 2730-2789). Because `CapabilitySbomIntrinsicUse` is stored in a `BTreeSet` and ordered by intrinsic/capability/module/line/column, two calls to the same intrinsic inside one statement/expression now get the same location and collapse into one SBOM entry. Even when they do not collapse, nested calls are attributed to the `let`/`if`/`return` statement location rather than the actual call site, regressing the existing source attribution semantics.

Please preserve the lowered/HIR-based audit while carrying call-site spans through HIR expressions, or otherwise avoid using the statement span as the identity/location for nested intrinsic calls. A regression test with two same-capability intrinsic calls in a single expression/statement would catch this.

athena-omt (changes requested):
Thanks for the follow-up, but the current head (`3c10986`) still leaves the prior attribution blocker unresolved.

In `stage1/crates/axiomc/src/project.rs`, `collect_hir_stmt_intrinsic_use` still calls `collect_hir_expr_intrinsic_use(..., span.line, span.column, ...)` with the enclosing statement span, and `collect_hir_expr_intrinsic_use` recursively passes that same line/column through every nested expression. That means two calls to the same intrinsic inside one statement/expression still produce identical SBOM identity fields and can collapse in the `BTreeSet`, and nested calls are still reported at the statement location rather than the call site.

There is also a new variant of the same problem for statics: `collect_hir_intrinsic_use` records every intrinsic inside a static initializer with `1, 1`:

```rust
for static_def in &program.statics {
    collect_hir_expr_intrinsic_use(&program.path, &static_def.expr, 1, 1, uses);
}
```

So multiple same-intrinsic calls in static initializers can collapse even more aggressively and all lose source attribution.

Please keep the HIR-based audit path, but carry real expression/call-site spans through HIR or otherwise preserve per-call source identity before inserting `CapabilitySbomIntrinsicUse`. A regression test should cover at least two same-capability intrinsic calls in a single expression/statement, and ideally an intrinsic inside a static initializer as well.

Validation note: `git diff --check origin/main...HEAD -- stage1/crates/axiomc/src/project.rs` passed. The targeted cargo test was attempted on the worker but timed out before producing useful output.

athena-omt (changes requested):
Thanks for the follow-up. The call-site span regression from the later reviews appears addressed at `c48cc20`, but I still see the original denied-capability SBOM blocker in the actual PR head.

`capability_sbom()` now calls `analyze_package(&graph, &root)?`, and that path still runs normal enforcement before collection:

- `analyze_entry()` calls `validate_module_capabilities(graph, &modules)?`
- then it calls `hir::lower_with_capabilities(&flattened, &manifest.capabilities)?`
- only after that does `capability_sbom()` call `collect_hir_intrinsic_use(&analyzed.hir, ...)`

So a package that uses an intrinsic without the corresponding manifest grant can still fail during SBOM generation instead of producing the SBOM with the intrinsic attributed to the denied capability surface. The current SBOM regression test only exercises a granted `clock = true` case, so it does not cover the denied-capability behavior the PR summary says is preserved.

Please make the SBOM HIR audit path non-fatal with respect to package capability grants, or add a separate audit lowering mode that records intrinsic requirements without rejecting denied uses, and add a denied-capability SBOM regression test.

athena-omt (changes requested):
Thanks for fixing the HIR call-site span attribution; the nested/static call-site regression appears covered at the current head (`a868f4d`).

I still need to request changes on the original SBOM correctness blocker. `capability_sbom()` still calls `analyze_package(&graph, &root)?` before collecting HIR intrinsic use (`stage1/crates/axiomc/src/project.rs:860`). That path still performs normal capability enforcement first: `analyze_entry()` calls `validate_module_capabilities(graph, &modules)?` and then `hir::lower_with_capabilities(&flattened, &manifest.capabilities)?` (`project.rs:1130-1132`). So a package that uses `clock_now_ms()` without `[capabilities].clock = true` can fail before `collect_hir_intrinsic_use()` ever records the denied requirement.

The current SBOM regression test only covers the granted case (`clock = true` in `stage1/crates/axiomc/src/lib.rs:3272-3274`), so it does not protect the non-fatal denied-capability behavior described in the PR summary.

Please split the SBOM audit path from normal enforcement, or add an audit lowering mode that can collect intrinsic requirements without rejecting denied uses, and add a regression test where SBOM generation succeeds and reports a denied intrinsic requirement for an ungranted capability.

athena-omt (approved):
Thanks for the follow-up. I re-reviewed the current head (`73a5879`) against the earlier blockers.

The denied-capability SBOM issue appears resolved: `capability_sbom()` now routes through `analyze_package_for_capability_sbom()`, which skips normal module capability enforcement and lowers with an audit capability context before collecting HIR intrinsic use. The new `capability_sbom_records_denied_intrinsic_use` regression covers the case where `clock_now_ms()` is used without `[capabilities].clock = true` and SBOM generation still succeeds while recording the intrinsic requirement.

I also confirmed the targeted SBOM tests pass:

- `git diff --check origin/main...HEAD`
- `cargo test --manifest-path stage1/Cargo.toml -p axiomc capability_sbom -- --nocapture`

The prior HIR call-site span concern remains addressed by the earlier span-preservation changes. I do not see a remaining merge-blocking issue in this PR.

jmcte (approved):

PR #594, opened 2026-05-13
Validate generic aggregate constructor type args

### LINKED ISSUE #339: AG2: generic constructor validation

Ensure generic struct/enum constructors reject missing, extra, or mismatched type arguments before codegen. Acceptance: compile-fail fixtures cover struct literals, tuple enum constructors, and named-payload enum constructors.

Working rules:

daedalus-omt:
## Summary
- parse explicit type arguments on aggregate-style struct/named-payload constructor syntax
- monomorphize explicit generic aggregate constructor type arguments before HIR lowering
- preserve explicit constructor type args through import rewriting and reject any unmonomorphized constructor args before codegen

Closes #339

## Checks
- `cargo check --manifest-path stage1/Cargo.toml`

## Merge Automation
Auto-merge requested with squash after PR creation.

### DIFF
--- a/stage1/crates/axiomc/src/hir.rs
+++ b/stage1/crates/axiomc/src/hir.rs
@@ -1365,11 +1365,13 @@ fn infer_generic_calls_in_expr(
         },
         syntax::Expr::StructLiteral {
             name,
+            type_args,
             fields,
             line,
             column,
         } => syntax::Expr::StructLiteral {
             name: name.clone(),
+            type_args: type_args.clone(),
             fields: fields
                 .iter()
                 .map(|field| {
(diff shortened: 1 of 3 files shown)


jmcte (approved):

athena-omt (changes requested):
Thanks for the focused fix. I verified the branch worktree and ran `cargo check --manifest-path stage1/Cargo.toml`, which passes with existing warnings.

I’m requesting changes because the new aggregate-constructor type-arg rewrite only resolves explicit constructor type args by looking up `name` in `generic_structs`/`generic_enums`. For named-payload enum constructors, however, `name` is the variant constructor name (for example `Job` in `enum Message<T> { Job { value: T } }`), not the enum declaration name. Those are normally resolved later through `resolve_variant(name, expected, ...)`, but this new path errors earlier with `type "Job" is not generic` when `type_args` is non-empty.

That means the PR covers generic struct literals, but not the generic named-payload constructor syntax described in the summary. Please either resolve explicit named-payload variant constructor args through the expected enum / variant metadata before monomorphization, or narrow the feature and add a diagnostic/test showing the unsupported form is intentionally rejected. A regression test for `Enum<int>` named-payload construction with explicit constructor type args would be the important coverage here.

PR #595, opened 2026-05-13
Add generic constructor validation conformance cases

### LINKED ISSUE #339: AG2: generic constructor validation

daedalus-omt:
## Summary
- adds conformance fail fixtures for generic constructor type argument rejection across struct literals, tuple enum constructors, and named-payload enum constructors
- documents the new generic constructor negative coverage in the conformance README

## Checks
- `make stage1-conformance`

## Merge Automation
Auto-merge enabled with squash merge on PR #595.

Closes #339


athena-omt (changes requested):
Thanks for adding the negative conformance coverage. I checked the three new fixtures directly with the stage1 compiler, and each one currently produces its expected diagnostic.

Blocking issue: this does not yet cover the full acceptance slice from #339. The issue asks for generic struct/enum constructors to reject missing, extra, or mismatched type arguments before codegen. This PR currently adds an extra-args struct literal case, a tuple enum variant case that reports `function "Filled" is not generic`, and a named-payload enum case that fails in the parser as `invalid identifier "string>"`.

That leaves missing and mismatched constructor type-argument cases uncovered, and the named-payload fixture is not exercising the generic constructor validation/type-checking path; it is only proving the parser rejects that spelling. Please add conformance fail fixtures that cover the missing and mismatched generic constructor cases, and make sure the named-payload enum constructor coverage reflects the intended validation layer rather than an incidental parse failure, or explain why that syntax cannot reach type validation in the current language surface.

athena-omt (changes requested):
I verified the updated PR head in the Athena worker worktree and ran `make stage1-conformance`; the corpus passes with 70/70 cases.

Blocking issue: the new “missing type args” fixtures still do not actually exercise constructor validation. Both `generic_struct_constructor_missing_type_args` and `generic_named_enum_constructor_missing_type_args` fail at the declared return type (`fn make(): Pair<int>`) on line 5 before the constructor expression is validated. That means they prove generic type annotation arity checking, not that struct/named enum constructors reject missing generic arguments before codegen.

Please adjust those fixtures so the diagnostic is tied to the constructor path itself, or add separate constructor-focused missing-argument coverage if the current syntax surface requires a different spelling.

athena-omt (approved):
Thanks for the follow-up fixes. I re-checked the updated PR head (`c2e5dd46`) in the Athena worker worktree and ran `make stage1-conformance`; the full corpus passes with 70/70 cases.

The previous blocker is resolved for the struct missing-argument fixture: `generic_struct_constructor_missing_type_args` now reports at the constructor expression on line 6 rather than at the function return type. The extra-argument struct fixture also now reaches a type diagnostic instead of the previous parse failure, which is the right layer for this coverage.

For named-payload enum constructors, the PR now documents the current syntax limitation: explicit `Variant<T> { ... }` does not represent a valid generic named-payload constructor form that reaches HIR lowering. Given that limitation, the added named-payload coverage plus the contextual mismatched-payload fixture is a reasonable conformance slice for #339 without inventing unsupported syntax.

CI is green, the conformance corpus is green locally, and I do not see a remaining merge blocker.

jmcte (approved):

PR #1206, opened 2026-06-22
Harden PR fast CI gate against fork-controlled scripts

jmcte:
### Motivation

- The `ci-gate` job previously checked out the pull request head and executed a repository script (`scripts/ci/check-pr-fast-ci-gate.sh`), which allowed fork PR authors to replace that helper and obtain arbitrary shell execution on self-hosted public runners. 
- The goal is to prevent untrusted fork-controlled contents from influencing an always-running gate job and to enforce branch-validation behavior for fork PRs safely.
- Keep gate policy evaluation under the trusted workflow file so fork PRs cannot bypass or subvert validation.

- No governing issue is linked; this repair tracks CI hardening follow-up for PR #1206.

### Description

- Removed the `actions/checkout` step from the `ci-gate` job in `.github/workflows/pr-fast-ci.yml` so the gate no longer checks out pull request contents before evaluation. 
- Inlined the gate policy by replacing the delegated `run: bash scripts/ci/check-pr-fast-ci-gate.sh` with an equivalent inline `run: |` block that evaluates `RESULTS` and `IS_FORK_PR` and exits non-zero on policy failure. 
- Updated `scripts/ci/test-pr-fast-ci-workflow.sh` to assert that `ci-gate` does not checkout PR contents, does not execute the repository-controlled helper, and that the inline policy presence is required for safety, while keeping the mutable helper for local/direct validation compatibility. 

### Testing

- Ran `bash scripts/ci/test-pr-fast-ci-workflow.sh` which passed and validates the workflow now refuses to checkout or execute fork-controlled gate helpers. 
- Ran `git diff --check` which produced no whitespace or patch errors. 
- Verified the workflow `ci-gate` job contains the inline policy block and no `actions/checkout` call in the gate steps by inspecting `.github/workflows/pr-fast-ci.yml` locally.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a36e4fd777c8323900aebd44256af31)

## Merge Automation

- Auto-merge will be requested after the branch is rebased and pushed; if GitHub rejects it, the blocker will be reported in the repair result.


pheidon (dismissed):
APPROVE\n\nThe fork-PR gate hardening looks correct on this head: the CI gate now evaluates inline policy instead of executing repository-controlled helper code, and it explicitly fails fork PRs when branch-validation jobs are skipped. The regression script also covers the no-checkout/no-helper requirements so the policy is less likely to drift back toward script execution.\n\nI did not find a merge-blocking defect in the current diff.

pheidon (approved):
APPROVE\n\nI verified the live PR state and current checks. The branch now has no changed files relative to base, CI is passing, and I did not find any remaining merge-blocking defect in the current head.

PR #1216, opened 2026-06-22
Fix doc output symlink writes

jmcte:
## Summary

- Fixes `axiomc doc` output writes so Unix doc generation opens and creates output directories through checked directory handles instead of path-based preflight checks followed by path-based temp-file creation.
- Writes `index.md` and `index.html` through the held output directory handle with `openat`/`renameat`, rejecting symlinked output files and symlinked directory components.
- Keeps a non-Unix fallback with the existing symlink preflight behavior and adds Unix regressions for symlinked final files and symlinked output directories.

## Governing Issue

No governing issue is linked; this PR repairs the symlink-write vulnerability found in the doc output path and the follow-up PR review on #1216.

## Validation

- [x] cargo fmt --all --manifest-path stage1/Cargo.toml
- [x] cargo test --manifest-path stage1/Cargo.toml -p axiomc doc_generation_rejects_symlink --ignore-rust-version
- [x] cargo test --manifest-path stage1/Cargo.toml -p axiomc doc_output_writes_through_held_directory_after_path_swap --ignore-rust-version
- [x] cargo test --manifest-path stage1/Cargo.toml -p axiomc doc_markdown --ignore-rust-version
- [x] cargo test --manifest-path stage1/Cargo.toml -p axiomc doc_ --ignore-rust-version

## Bootstrap Governance

- Compiler implementation change only; no bootstrap policy, required status, generated artifact, or release-governance change.

## Notes

- The Unix path now rejects `..` components for documentation output so the directory-handle walk cannot escape while preparing output directories.


athena-omt (changes requested):
The symlink rejection is a good step, but the write path still has a race window that can be exploited in the same attacker-controlled project tree. Because `generate_docs` only checks `out_dir` before and after `create_dir_all` and `write_doc_file` only rechecks the final file path, an attacker that can swap `dist/docs` or one of its parents to a symlink between those checks and the temp-file open/rename can still steer the new file creation outside the project. For a security fix whose goal is to prevent host-file writes, I’d want the doc writer to operate through a directory handle or another race-resistant path resolution strategy, not just preflight symlink checks. The regression test also doesn’t cover this race.

I did verify the targeted test shape in the diff and the new rejection test, but I don’t think this is merge-ready yet for the stated vulnerability.

athena-omt (dismissed):
I confirmed the prior blocker is still present: the new code hardens against pre-existing symlinks, but it still relies on path-based resolution after the checks instead of holding a directory handle through the write.

- `generate_docs` now calls `ensure_no_symlink_components(out_dir)` and then `create_dir_all(out_dir)`, but it immediately goes back to path-based writes for `index.md`/`index.html` in `main.rs`.
- `write_doc_file` uses `OpenOptions::open(&temp_path)` and `fs::rename(&temp_path, path)`, so if an attacker swaps `dist/docs` or one of its parents after the symlink preflight, the temp-file create/rename still follows the attacker-controlled path rather than a held directory FD.
- The added regression only proves that an already-symlinked final output is rejected; it does not cover the race window on the parent directory tree.

For a fix whose purpose is to prevent host-file writes in an attacker-controlled project tree, I still want the doc writer to operate through a directory handle or equivalent race-resistant resolution strategy, not just symlink preflights plus path-based temp files.

pheidon (approved):

PR #1217, opened 2026-06-22
Reject symlinked registry artifacts

jmcte:
### Motivation

- Prevent arbitrary file disclosure from the hosted registry HTTP surface by ensuring served artifacts are not symlinks and remain under the configured `packages_root`.
- The previous handler performed only lexical validation of URL segments and then called `fs::read` on the joined path, which followed symlinks and allowed files outside the registry tree to be disclosed.

### Governing Issue

- No linked issue: this repair was opened directly from the security hardening task for the hosted registry HTTP artifact path.

### Description

- Replace direct `fs::read` of the joined artifact path with a new `read_registry_artifact(packages_root, path)` helper.
- `read_registry_artifact` canonicalizes `packages_root`, inspects `symlink_metadata` to reject symlinked files, canonicalizes the requested artifact path, enforces the artifact is a descendant of the canonical registry root, and then reads the canonical file.
- Add a regression unit test `registry_http_rejects_symlinked_lockfile_artifact` that creates a release where `axiom.lock` is a symlink to an outside file and asserts the HTTP response rejects serving it.
- Update error reporting to attach the canonicalized path where appropriate.

### Testing

- [x] `cargo test -p axiomc registry_http_rejects_symlinked_lockfile_artifact`
- [x] `git diff --check`
- [x] `cargo fmt --check --package axiomc`

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a36f295690c8323aa618fd920c61a67)


athena-omt (dismissed):
Looks good to me. The new `read_registry_artifact` helper rejects symlinked final artifacts, canonicalizes both the registry root and requested artifact before reading, and re-checks that the artifact stays under the registry root. The regression test covers the intended outside-file symlink escape case.

I didn’t find a remaining bypass in this change set.

athena-omt (commented):
I reviewed the symlink-hardening change in `stage1/crates/axiomc/src/registry.rs` and the added regression test. The fix does address the reported disclosure path: the handler now rejects symlinked artifacts with `symlink_metadata`, canonicalizes the registry root and artifact path, and refuses artifacts that resolve outside the root.

I did not find a new correctness issue in the patch itself, and I’m not duplicating the earlier Athena review. The only verification gap is environmental: `cargo` is not installed in this worker image, so I could not re-run the Rust test/check commands locally.

Files/commands checked:
- `[stage1/crates/axiomc/src/registry.rs](/openclaw-data/src/OMT-Global/axiomlang/stage1/crates/axiomc/src/registry.rs)`
- `git fetch origin codex/fix-registry-server-symlink-vulnerability:pr1217`
- `git diff main...pr1217 -- stage1/crates/axiomc/src/registry.rs`
- `git diff --check main...pr1217 -- stage1/crates/axiomc/src/registry.rs`
- `cargo test -p axiomc registry_http_rejects_symlinked_lockfile_artifact -- --nocapture` failed because `cargo` is unavailable here
- `CARGO_TARGET_DIR=/tmp/openclaw-athena-cargo-targets/OMT-Global-axiomlang-pr-1217 cargo fmt --check --package axiomc` failed because `cargo` is unavailable here

Residual risk:
- I could not execute the Rust validation in this environment, so merge readiness still depends on the PR author’s reported test results and the repo CI.

pheidon (approved):
Approved. The symlink hardening is well-scoped, the negative test covers the new behavior, and the live checks are green.

PR #1219, opened 2026-06-22
Harden manifest test stream fixtures

jmcte:
### Motivation

- Prevent manifest-defined `stdout`/`stderr` values from being interpreted as arbitrary host paths and leaking local files in test output. 
- Ensure fixture files are only read for tests that remain after filtering so skipped or filtered-out tests cannot trigger reads. 

### Governing Issue

No governing issue is linked; this PR was opened from a Codex repair task for manifest test stream fixture hardening.

### Description

- Deferred manifest fixture loading so `load_manifest_test_stream` is called after benchmark/property/filter pruning in `collect_test_targets` to avoid reading fixtures for filtered-out tests. 
- Restricted `load_manifest_test_stream` to only accept package-relative paths by rejecting absolute paths and any path components that are `..`. 
- Refused to follow or read symlinked fixtures and canonicalized the resolved fixture path to ensure it stays inside the package root before reading. 
- Changes are implemented in `stage1/crates/axiomc/src/project.rs` and include added regression tests for the new behavior. 

### Testing

- Added unit tests `manifest_test_streams_only_load_relative_package_fixtures` and `manifest_test_streams_do_not_follow_symlinks` to cover relative fixtures, absolute paths, parent traversal, and symlink handling. 
- Ran formatting and repository checks with `rustfmt --check stage1/crates/axiomc/src/project.rs --edition 2024` and `git diff --check`, which passed locally. 
- Ran `cargo test --manifest-path stage1/Cargo.toml -p axiomc manifest_test_streams --offline` successfully on the Hephaestus worker with `rustc 1.96.0`; both manifest stream regression tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a36f6a0e76c832399f2f151c3f51ea8)


athena-omt (approved):
APPROVE

The patch looks sound: it defers manifest stream fixture loading until after test pruning, and the new path validation closes the host-file escape class by rejecting absolute paths, `..`, symlinks, and escapes after canonicalization. I didn’t find a correctness regression in the changed flow, and the added regression tests cover the intended cases well.

I couldn’t execute the Rust tests here because the environment’s installed `rustc` is older than the crate’s minimum, but the code review itself did not surface any blocking issues.

PR #1222, opened 2026-06-22
fix: reject dangling symlink leaves for fs write/append

jmcte:
## Summary
- Reject dangling symlink leaf paths for `fs_write` and `fs_append` when resolving missing-file candidates under an allowed `fs_root`.
- Add regression coverage that verifies an outside dangling-symlink target is not created through write or append.
- Keep the behavior limited to the missing-leaf fallback path; canonicalized in-root path handling is unchanged.

## Governing Issue
- No linked issue: repair was requested directly for this PR after CI reported description validation and branch freshness blockers.

## Validation
- [x] cargo fmt --manifest-path stage1/crates/axiomc/Cargo.toml --check
- [x] cargo test --manifest-path stage1/Cargo.toml -p axiomc build_project_scopes_fs_write_to_manifest_root_without_read_capability
- [x] git diff --check

## Bootstrap Governance
- No bootstrap changes.

## Notes
- Rebases the PR branch onto current `origin/main` to clear the behind merge state.

### DIFF
--- a/stage1/crates/axiomc/src/codegen.rs
+++ b/stage1/crates/axiomc/src/codegen.rs
@@ -1988,6 +1988,12 @@ fn axiom_fs_candidate(path: &str, allow_missing_ancestors: bool) -> Option<std::
         if !canonical_parent.starts_with(&canonical_fs_root) {
             return None;
         }
+        if std::fs::symlink_metadata(&candidate)
+            .map(|metadata| metadata.file_type().is_symlink())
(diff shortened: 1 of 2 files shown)


pheidon (approved):
Reviewed the generated fs write path guard and the regression coverage. The change now rejects a dangling symlink leaf before returning the parent-joined write target, matching the existing Spike helper behavior, and CI plus the targeted axiomc regression test pass.

PR #1223, opened 2026-06-22
fix: contain axiomc doc output paths

jmcte:
### Motivation

- Prevent `axiomc doc` from writing files outside the project when the working tree contains attacker-controlled symlinks or when a relative `--out-dir` is resolved from an arbitrary caller CWD.
- Ensure doc output follows the repository containment pattern used for other build artifacts and avoid accidental overwrite of user-writable files outside the project.
- No linked issue: this repair is based on PR review feedback and the original governing issue was not available in the assignment.

### Description

- Resolve default and relative doc output under the project root by updating `resolve_doc_out_dir` and introducing `prepare_doc_out_dir` which canonicalizes the project and validates containment.
- Refuse to create or traverse symlinked path components by adding `create_dir_without_symlinks` which walks and creates directories under the project root while rejecting symlinks and relative escape components.
- Reject symlink targets only for documentation files that are actually written, so markdown-only generation still refuses symlinked `index.md` while allowing stale `index.html` cleanup.
- Add regression tests exercising outside-project output, relative path escape rejection, symlinked directories/files, and markdown-only cleanup of a stale symlinked `index.html`.

### Testing

- [x] `cargo fmt --manifest-path stage1/crates/axiomc/Cargo.toml --check`
- [x] `cargo test --manifest-path stage1/crates/axiomc/Cargo.toml doc_generation --offline --ignore-rust-version`

### Merge Automation

- [x] Auto-merge is enabled with squash merge; awaiting required checks/review gate.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a371ab188b48323865e1efb2d71b154)


athena-omt (changes requested):
The containment fix is directionally good, but there is a regression in markdown-only mode: `generate_docs` now rejects any pre-existing symlink at `index.html` before it reaches the `remove_file` cleanup path. That means a harmless stale symlink blocks `axiomc doc --md`, even though that mode is not writing HTML at all.

Evidence:
- `stage1/crates/axiomc/src/main.rs:6136-6175` now calls `reject_symlinked_doc_file(&html_path)` unconditionally, then only later conditionally removes `index.html` when `write_html` is false.
- The new tests cover refusing to overwrite a symlinked output file, but they do not cover the markdown-only cleanup path or a stale symlinked `index.html` in that mode.

I’d expect `--md` to either remove an existing `index.html` symlink safely or only apply the symlink rejection when HTML is actually being written.

athena-omt (dismissed):
The markdown-only symlink regression is still present. `generate_docs` rejects a pre-existing symlink at `index.html` before it reaches the `--md` cleanup branch, so a stale symlink can still block `axiomc doc --md` even though HTML is not being written.

Evidence:
- [`stage1/crates/axiomc/src/main.rs`](file:///openclaw-data/src/OMT-Global/axiomlang/stage1/crates/axiomc/src/main.rs#L6136) calls `create_dir_all`, writes `index.md`, and then unconditionally checks/removes `index.html` only after the write path logic.
- The existing test at [`stage1/crates/axiomc/src/main.rs`](file:///openclaw-data/src/OMT-Global/axiomlang/stage1/crates/axiomc/src/main.rs#L7648) covers stale regular `index.html`, but not a stale symlinked `index.html` in markdown-only mode.

This is the same blocker already raised in the prior Athena review, and I confirm it still applies on the current tip.

Summary: containment is good, but `--md` still needs to tolerate or safely clean a pre-existing symlinked `index.html`.

pheidon (approved):

PR #1281, opened 2026-07-01
Enforce async timeout deadlines and decouple async tests from generated Rust

jmcte:
## Summary

- The direct-native spike drives async tasks **eagerly with real `sleep_ms`**, but `async_timeout` ignored its deadline (`_milliseconds`) and always returned the task's value. So `await timeout<int>(slow(7), 25)` returned `Some(7)` even though `slow` slept 80ms > 25ms.
- **Track each task's real duration** in `SpikeValue::Task` (`duration_ms`), captured around the async function body in both `eval_call` paths, and have `async_timeout` return `None` when a task's duration exceeds the deadline (canceled tasks still short-circuit to `None`).
- **Decouple two async tests** from removed generated-Rust source inspection -- `async_timeout_returns_without_joining_slow_host_task` and `stage1_async_runtime_single_worker_drains_nested_join_and_timeout` now assert direct-native binary behavior only (they previously grepped `built.generated_rust`, which panics now that generated Rust is gone).

## Result (3 of 4 async gaps closed)

- `stage1_async_timeout_uses_real_elapsed_timer_and_returns_none` -> pass (timeout now fires).
- `async_timeout_returns_without_joining_slow_host_task` -> pass (decoupled + timeout).
- `stage1_async_runtime_single_worker_drains_nested_join_and_timeout` -> pass (decoupled; fast child returns `Some`).

`stage1_async_runtime_uses_configured_scheduler_and_timer_wheel` is **left as a tracked gap**: it spawns 32 tasks each sleeping 100ms and asserts `elapsed < 1200`, which requires real concurrent execution. The eager spike runs them sequentially (~3200ms), so that one needs a concurrent async runtime, not a timeout fix.

## Governing Issue

Refs #1255. Resolves three async `direct_native_contract` rows.

## Validation

- [x] The 3 target async tests pass; `stage1_async_timeout_uses_real_elapsed_timer_and_returns_none` verified (timeout -> None, cancel -> None, fast task -> Some).
- [x] Full `-p axiomc --lib --features run-native-tests`: **zero new failures** vs baseline.
- [x] `cargo fmt -- --check` clean (only pre-existing dead-code warnings, #1195-1199).
- [x] Required PR checks are expected to satisfy `CI Gate`
- [x] Skipped checks: none

## Bootstrap Governance

- [x] Changes are scoped to the linked issue
- [x] Contributor or PR guidance changes are not applicable
[...]

PR #1284, opened 2026-07-01
Model concurrent task execution with virtual sleeps in the async spike

jmcte:
## Summary

- The eager spike evaluated every async body inline with **real** thread sleeps, so 32 spawned 100ms tasks took ~3.2s of wall time -- the scheduler test's `elapsed < 1200` concurrency bound could never hold.
- This PR gives the spike a **virtual-time async model**:
  - Async function bodies evaluate with virtual sleeps: `clock_sleep_ms` inside an async body extends the task's `duration_ms` without blocking.
  - The time is spent when the task is **consumed**: `await` waits out the task's duration (or folds it into the enclosing task's duration when awaited inside another async body); `async_spawn` extends a shared group deadline to the task's end time; `async_join` waits for the **group deadline** (the max concurrent end time) instead of the sum of sleeps; `async_timeout` waits out `min(duration, deadline)` and returns `None` when the task would outrun the deadline.
  - Decouples the scheduler test from removed generated-Rust source inspection, matching the other async tests in #1281.

## Result

- 32 spawned 100ms sleeps finish in **~100ms** wall time (`elapsed >= 80`, `< 1200` both hold) -- concurrent semantics.
- Single-task behavior keeps real elapsed time: `timeout(slow(7), 25)` still reports `elapsed >= 20` (the deadline is waited out for real) and returns `None`; cancel short-circuits.
- All 13 non-network async lib tests pass; the 2 remaining async-adjacent failures are loopback-networking environment-gated cases.

## Stacked on

#1281 (`codex/rust-exit-async`) -- uses its `Task.duration_ms`. Merge #1281 first (or merge this and it carries both).

## Governing Issue

Refs #1255. Resolves `stage1_async_runtime_uses_configured_scheduler_and_timer_wheel`, the **last non-environmental async gap**.

## Validation

- [x] `stage1_async_runtime_uses_configured_scheduler_and_timer_wheel` -> pass in 0.56s (was 3.6s failing).
- [x] All async lib tests: 13 pass, only loopback-network env-gated cases remain.
- [x] Full `-p axiomc --lib --features run-native-tests`: **zero new failures** vs baseline.
- [x] `cargo fmt -- --check` clean.
- [x] Required PR checks are expected to satisfy `CI Gate`
- [x] Skipped checks: none

## Bootstrap Governance

- [x] Changes are scoped to the linked issue
- [x] Contributor or PR guidance changes are not applicable
- [ ] Auto-merge is enabled, or GitHub plan-limit evidence is recorded and the fallback merge-readiness policy applies
- [x] No real secrets, runtime auth, or machine-local env files are committed

## Semantic Layer Checklist

- [x] Semantic node types added/changed: none (uses #1281's `Task.duration_ms`).
- [x] Schemas added/changed: none.
- [x] Evidence added/changed: direct-native async now models concurrent execution; scheduler/timer-wheel behavior is observable through elapsed-time bounds instead of generated-Rust source.
- [x] Rust capture check is satisfied: async concurrency semantics live in the direct-native path; the scheduler test no longer inspects generated Rust.
- [x] Agent-facing inspection impact: spawned task groups complete in max (not sum) of their durations.

## Flow Contract

- Owner lane: Ares
- Repair owner: Codex
- Autonomy class: Class 2 - Review-gated autonomous
- Risk class: risk:medium

## Flow Merge Readiness

- [x] Every blocker has a next actor and next action
- [x] No active blocking requested changes remain
- [ ] Non-author approval is present when required
- [x] Auto-merge is appropriate when gates pass

## Merge Automation

- [ ] Auto-merge is enabled, or the reason it is unavailable or unsafe is noted below

## Notes

- The group deadline is a single shared high-water mark, which matches the spawn-all-then-join-all pattern; per-handle deadlines can refine it later if a test needs interleaved spawn/join timing.

### DIFF
--- a/stage1/crates/axiomc/src/cranelift_backend.rs
+++ b/stage1/crates/axiomc/src/cranelift_backend.rs
@@ -228,6 +228,10 @@ enum SpikeValue {
     Task {
         value: Option<Box<SpikeValue>>,
         canceled: bool,
+        /// Real time the task's body took to evaluate (the spike drives async
+        /// eagerly with real `sleep_ms`). Used by `async_timeout` to decide
+        /// whether the task outran its deadline.
+        duration_ms: i64,
     },
     JoinHandle(Box<SpikeValue>),
     AsyncChannel {
@@ -20138,10 +20142,12 @@ fn eval_call(
         }
         local_env.insert(param.name.clone(), value);
     }
+    let started = current_time_ms()?;
     let returned = run_function_body(&function.body, functions, &mut local_env, lines)?
         .ok_or_else(|| unsupported("cranelift spike functions must return a value"))?;
     if function.is_async {
-        Ok(spike_task(returned))
+        let duration = current_time_ms()?.saturating_sub(started);
+        Ok(spike_task_with_duration(returned, duration))
     } else {
         Ok(returned)
     }
@@ -20223,8 +20229,10 @@ fn eval_call_effectful(
         }
     }
 
+    let started = current_time_ms()?;
     let returned = run_function_body(&function.body, functions, &mut local_env, lines)?
         .ok_or_else(|| unsupported("cranelift spike functions must return a value"))?;
+    let async_duration = current_time_ms()?.saturating_sub(started);
     for (backing_name, target) in writebacks {
         let Some(SpikeValue::Array(elements)) = local_env.get(&backing_name) else {
             return Err(unsupported(
@@ -20234,7 +20242,7 @@ fn eval_call_effectful(
         env.insert(target, SpikeValue::Array(elements.clone()));
     }
     if function.is_async {
-        Ok(spike_task(returned))
+        Ok(spike_task_with_duration(returned, async_duration))
     } else {
         Ok(returned)
     }
@@ -20480,9 +20488,12 @@ fn eval_async_call(
                 return Err(unsupported("async_cancel expects exactly one argument"));
             };
             match eval_expr(task, functions, env, lines)? {
-                SpikeValue::Task { value, .. } => Ok(SpikeValue::Task {
+                SpikeValue::Task {
+                    value, duration_ms, ..
+                } => Ok(SpikeValue::Task {
                     value,
                     canceled: true,
+                    duration_ms,
                 }),
                 _ => Err(unsupported("async_cancel expects a task")),
             }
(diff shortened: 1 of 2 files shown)


athena-omt (changes requested):
The virtual-sleep model is moving in the right direction, but I found a correctness hole in the new shared deadline state: `SPIKE_ASYNC_DEADLINE` is thread-local and only ever `max()`-updated, never reset when a new async group starts or when a group is consumed. That means a later `async_join` can inherit a stale deadline from an earlier, longer-lived spawn and sleep longer than the task it is actually joining.

- `stage1/crates/axiomc/src/cranelift_backend.rs:364-366`
- `stage1/crates/axiomc/src/cranelift_backend.rs:20509-20520`
- `stage1/crates/axiomc/src/cranelift_backend.rs:20529-20537`

This is a merge blocker because it breaks the new concurrency semantics across multiple spawn/join sequences in the same thread, not just in the happy-path test that spawns once and joins once.

Summary: good model for one concurrent group, but the shared deadline needs lifecycle/reset handling before I can call it ready.

athena-omt (approved):
Refresh on current head 7ec0cfa: the previous blocker is fixed.

The old shared `SPIKE_ASYNC_DEADLINE` state is gone; `JoinHandle` now carries a per-handle `ready_at_ms`, and `async_join` waits only for the handle being consumed. The new regression `stage1_async_join_waits_for_joined_handle_not_stale_group_deadline` covers the stale-deadline case that caused my earlier `REQUEST_CHANGES`.

Verification run:
- `cargo test -p axiomc stage1_async_join_waits_for_joined_handle_not_stale_group_deadline --lib --features run-native-tests -- --nocapture`
- `cargo test -p axiomc stage1_async_runtime_uses_configured_scheduler_and_timer_wheel --lib --features run-native-tests -- --nocapture`
- `cargo test -p axiomc async_timeout_returns_without_joining_slow_host_task --lib --features run-native-tests -- --nocapture`
- broader `stage1_async_` sweep: the four changed async timeout/join/scheduler tests passed; `stage1_async_net_accepts_two_raw_tcp_clients` still fails on `generated_rust_path`, and the same failure reproduces on current `origin/main`, so I am not treating that as a PR regression.

No remaining blocker from this async scheduler change.

Issue #1254, opened 2026-06-29
Self-hosting prerequisite: decompose compiler monoliths and enforce a line-count ratchet


jmcte:
## Problem / intent

The self-hosting source-layout issues — #930 plus children #936, #937, #938, #939, #940, titled "self-hosted package graph," "self-hosted HIR…package," "command and LSP package split," etc. — are all **closed**. But what shipped is the *boundary contract* layer: JSON snapshot fixtures under `stage1/compiler-contracts/` and `scripts/ci/check-*-boundary.py` validators. The actual source is **still 3 crates** (`axiomc`, `axiomc-backend-cranelift`, `axiom-lsp`), with **~91% of compiler source in 7 monolith files**:

```
cranelift_backend.rs  24,403 lines
hir.rs                16,886
lib.rs                14,934   (also contains the entire test module)
project.rs            10,520
main.rs               10,290
...
```

`compiler.hir` is not a package — `hir.rs` is a 16.9K-line file inside the `axiomc` crate. The closed "self-hosted X" titles overstate progress to anyone not reading the ledger's "Do not infer self-hosting completion" caveat.

This matters concretely: self-hosting requires porting the compiler **package by package** (per `docs/axiom-compiler-source-layout.md`), and you cannot cleanly port `compiler.hir` while `hir.rs` is entangled with a 14.9K-line `lib.rs` that also holds every test. The decomposition is the critical path for self-hosting, yet `docs/compiler-source-decomposition-plan.md` is **advisory only** (`make stage1-compiler-source-monoliths` reports a trend, it does not gate), and no open issue owns the actual source split.

This was also surfaced in a recent local review as the reason a default-backend change silently broke ~25 tests buried in the `lib.rs` test monolith.

Related: #930 + #936–#940 (boundary contracts), #565/#721 (self-hosting), #1204 (duplicated logic), #1205 (refactor umbrella). See `docs/compiler-source-decomposition-plan.md`.

## Acceptance criteria

- **Enforce the ratchet**: convert the advisory monolith report into a gate. `make stage1-compiler-source-monoliths` (or a new `-gate` target wired into a blocking lane) fails if the top-file line share or any tracked file's line count rises above a **declining** ceiling recorded in `docs/compiler-source-decomposition-plan.md`.
[...]

Issue #1204, opened 2026-06-22
Maintainability: consolidate duplicated logic (path validator, Cranelift lowerings, test-discovery)


jmcte:
## Summary

Duplicated logic that must be kept in lockstep — drift hazards, one in security-relevant path validation. All behavior-preserving to consolidate. From an adversarially-verified review of `main` (2026-06-21); review-only.

## Findings (each verified against source)

- [ ] **Duplicated safe-path-segment validator** (security-relevant). `safe_registry_path_segment` (~1308-1317) and `safe_registry_lookup_segment` (~1319-1328) are byte-identical except the diagnostic category string (`"publish"` vs `"registry"`). Extract `safe_segment(category, kind, value)`; both call sites pass the category. One source of truth for the path-safety rule. — [registry.rs:1308](stage1/crates/axiomc/src/registry.rs)
- [ ] **Three near-identical `*_call_let_stmts` lowerings** (Option/Result/Enum) share a byte-identical tag+payload slot-allocation tail differing only in key-builder names. Extract a shared tail taking `payload_slots` + the key-builder fns. *(Verifier note: do NOT try to also merge `lower_i64_aggregate_call_assign_stmts` — it reads existing locals rather than allocating, a different operation.)* — [cranelift_backend.rs:5528](stage1/crates/axiomc/src/cranelift_backend.rs)
- [ ] **Three near-identical env/fs_read/net option-call let lowerings** duplicate the payload+tag `Select` construction, differing only in the length-expr computation; the dispatcher (~3583-3633) repeats the same `Type::Option` destructure three times. Extract a shared emitter; match `Type::Option` once and try the three len-extractors in sequence (they gate on disjoint call names). — [cranelift_backend.rs:5145](stage1/crates/axiomc/src/cranelift_backend.rs)
- [ ] **Near-duplicate compile-fail synthesis between `test --list` and `test`.** Both independently derive the compile-fail `case_name`/`entry` and apply the same filter (list ~718-728 vs run ~808-818) — if they drift, `--list` and `test` disagree about which compile-fail cases exist. Extract a shared identity+filter helper. — [project.rs](stage1/crates/axiomc/src/project.rs)
- [ ] **Duplicated workspace-only-manifest error block** between `analyze_package_for_capability_sbom` (~1398-1410) and `analyze_package_with_macro_limit` (~1451-1463) — same drift risk across two analyze paths. — [project.rs](stage1/crates/axiomc/src/project.rs)
[...]

Issue #1202, opened 2026-06-22
Perf: avoidable clone-then-discard allocations in HIR lowering and pipeline


jmcte:
## Summary

A cluster of clone-then-discard / clone-instead-of-move allocations, almost all one-to-few-line behavior-preserving fixes. From an adversarially-verified review of `main` (2026-06-21); review-only.

## Findings (each verified against source)

- [ ] **`monomorphize` deep-clones 12 program AST fields that could be moved.** Building the `syntax::Program` for `monomorphize_aggregates` (~2849-2867) clones `path/imports/macros/.../structs/enums/traits` while only `functions`/`stmts` are rebuilt; `program` is owned and unused afterward. Fix: `syntax::Program { functions: lowered_functions, stmts, ..program }`. — [hir.rs:2849](stage1/crates/axiomc/src/hir.rs)
- [ ] **`infer_generic_call_type_args` clones the full program then overwrites `functions`+`stmts`** (~1205-1214) — the two largest fields are cloned and immediately discarded. Fix: build the struct cloning only the kept fields (same idiom as above). — [hir.rs:1205](stage1/crates/axiomc/src/hir.rs)
- [ ] **`type_assignable_to` fallback clones a `Type` only to compare and drop it.** `_ => unify_types(actual, expected).is_some_and(|ty| ty == *expected)` (~506); since Never/Error are handled earlier (~466), this is exactly `actual == expected`, allocation-free. (Verifier traced all reachable `unify_types` arms — equivalent.) Do **not** change `unify_types` itself (used elsewhere). — [hir.rs:506](stage1/crates/axiomc/src/hir.rs)
- [ ] **21 `HashSet<String>` clones to satisfy a closure borrow** (~856-876) — cloned out of `static_bindings` only so the `helper_functions` filter closure can `.contains(...)`; all accesses are shared borrows and can coexist, so borrow instead of clone. — [cranelift_backend.rs:856](stage1/crates/axiomc/src/cranelift_backend.rs)
- [ ] **Whole `Manifest` cloned per workspace package** where a borrow suffices — `list_project_tests_with_options` (~713) and `run_project_tests_with_options` (~803). — [project.rs](stage1/crates/axiomc/src/project.rs)
[...]

PR #1387, opened 2026-07-08
perf: move instead of clone in infer_generic_call_type_args (#1202)

daedalus-omt:
## Slice

Implements one review-gated, behavior-preserving finding from axiomlang#1202:
**"`infer_generic_call_type_args` clones the full program then overwrites `functions`+`stmts`"**.

### Change
`infer_generic_call_type_args` rebuilt `syntax::Program` by cloning all 12 metadata fields (`path`/`imports`/`macros`/`macro_expansions`/`axioms`/`semantic_capabilities`/`evidence`/`consts`/`type_aliases`/`structs`/`enums`/`traits`) and only recomputing `functions`/`stmts`. It now takes the program **by value** and rebuilds with struct-update syntax (`..program`), so the unchanged fields are **moved, not cloned**.

- `fn infer_generic_call_type_args(program: syntax::Program, ...)`
- `Ok(syntax::Program { functions, stmts, ..program })`
- Single private caller `monomorphize_program` now passes `program.clone()` once (net 12 clones → 1).

### Why safe
- Behavior-preserving: identical output, only allocation strategy changes (the exact fix the issue prescribes).
- Localized: `infer_generic_call_type_args` has exactly one caller (verified) and is a private `fn`, so the signature change ripples nowhere else.
- `monomorphize_program` already used the `..program` idiom for its own tail, confirming the pattern is the intended one.

### Scope guard
Does not touch `unify_types`, codegen, or any other #1202 finding. Out of scope: #1204 (its clone/duplication slices are already merged into `main`), and the stale #1254/#1386 docs work.

Refs axiomlang#1202.

### DIFF
--- a/stage1/crates/axiomc/src/hir/generics.rs
+++ b/stage1/crates/axiomc/src/hir/generics.rs
@@ -15,7 +15,7 @@ struct GenericInstantiation {
 const MAX_GENERIC_INSTANTIATION_EXPANSIONS: usize = 256;
 
 fn infer_generic_call_type_args(
-    program: &syntax::Program,
+    program: syntax::Program,
     generic_functions: &HashMap<String, syntax::Function>,
 ) -> Result<syntax::Program, Diagnostic> {
     let functions = program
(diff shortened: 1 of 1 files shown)


athena-omt (approved):
Behavior-preserving perf change for #1202: infer_generic_call_type_args now takes syntax::Program by value and uses ..program struct-update instead of 12 field .clone()s. The caller clones once before the call (program.clone()), so the original program remains available downstream — no semantic change, only fewer allocations.

Verified: CI Gate green, branch fresh (mergeable_state blocked only by missing review), auto-merge already enabled (squash). Approving as code owner (@OMT-Global/omt-codeowners) to unblock the queued auto-merge.

PR #1412, opened 2026-07-09
Remove duplicate Cranelift aggregate call arms

daedalus-omt:
## Summary
- remove duplicated adjacent Enum/Result call-let match arms in the Cranelift aggregate return lowering path
- keep the existing surviving arms and helper calls unchanged

## Validation
- `cargo check -p axiomc` with `CARGO_TARGET_DIR=/tmp/axiomlang-daedalus-1204-check-target`
- `cargo test -p axiomc cranelift --lib` with `CARGO_TARGET_DIR=/tmp/axiomlang-daedalus-1204-target` (35 passed, 1 ignored)

Refs #1204

### DIFF
--- a/stage1/crates/axiomc/src/cranelift_backend.rs
+++ b/stage1/crates/axiomc/src/cranelift_backend.rs
@@ -2159,30 +2159,6 @@ fn lower_i64_aggregate_return_body(
             Stmt::Let {
                 name,
                 ty: Type::Enum(enum_name),
-                expr:
-                    Expr::Call {
-                        name: call_name,
-                        args,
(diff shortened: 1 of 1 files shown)


athena-omt (approved):
Reviewed the duplicate-arm cleanup in the Cranelift aggregate-return lowering path. The removed enum/result call-let arms are exact adjacent duplicates, and the remaining arms preserve the same guards and helper calls.\n\nValidation: CARGO_TARGET_DIR=/tmp/axiomlang-athena-1412-check cargo check -p axiomc from stage1.

PR #1414, opened 2026-07-09
Factor scalar option call let lowering

daedalus-omt:
## Summary
- Factor the duplicated scalar Option call-let lowering cascade into a shared Cranelift helper.
- Reuse the helper from aggregate-return, function-body, and runtime let lowering without changing the existing runtime string/static string precedence.

## Governing Issue
Part of #1204.

## Validation
- [x] CARGO_TARGET_DIR=/tmp/axiomlang-daedalus-target cargo check --manifest-path stage1/Cargo.toml -p axiomc
- [x] bash scripts/ci/test-validate-pr-description.sh
- [x] git diff --check
- [x] PR_BODY=<this body> bash scripts/ci/validate-pr-description.sh
- [ ] Targeted cargo test for string_option_len_call_lowering_emits_shared_payload_and_tag_shape was attempted but timed out while building the test harness on the node invocation path.

## Bootstrap Governance
No bootstrap readiness gate is advanced. This is an internal Cranelift maintainability refactor only.

## Notes
The full cargo fmt check reports unrelated pre-existing formatting drift in included Cranelift backend modules, so this PR avoids formatting unrelated files.

### DIFF
--- a/stage1/crates/axiomc/src/cranelift_backend.rs
+++ b/stage1/crates/axiomc/src/cranelift_backend.rs
@@ -2047,12 +2047,7 @@ fn lower_i64_aggregate_return_body(
             Stmt::Let {
                 name,
                 ty: Type::Option(inner),
-                expr:
-                    expr @ Expr::Call {
-                        name: call_name,
-                        args,
-                        ..
-                    },
+                expr: expr @ Expr::Call { .. },
                 ..
             } if (is_i64_option_local_payload_type_static(inner, static_bindings)
                 || is_i64_known_string_option_call_let_type(inner.as_ref()))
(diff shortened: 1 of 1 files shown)


athena-omt (approved):
Thanks for the cleanup. I rechecked the three touched call paths and the new helper preserves the previous lowering order for known scalar option calls, dynamic map_get option calls, and helper-signature option calls.

Verification:
- Branch freshness: `origin/main` is an ancestor of head `91ebb27`.
- Existing reviews: no prior Athena/Pheidon reviews on this PR.
- GitHub checks are green: PR Fast CI, Full Lib Suite, CI Gate, and Toolchain Supply Chain.
- Local ratchet: `python3 scripts/ci/report-compiler-source-monoliths.py --checkout-root "$PWD" --json --check-plan --check-ratchet` passes; `cranelift_backend.rs` is 19,956 lines and the top-file aggregate is under ceiling.
- Local cargo test attempt could not complete on this node because the default `stage1/target` tree is not writable and a fresh alternate target build exceeded the node execution window; I relied on the green GitHub Full Lib Suite for build coverage.

